PR #33, opened 2022-07-09
Bump version to 0.0.17

armstjc:


docs/docs/mlb/index.md
```diff
@@ -0,0 +1,374 @@
...
+
+### Example:
+
+```
+import sportsdataverse as sdv
+df = sdv.mlb.searchMlbPlayers(search="Votto",isActive="y")
+print(df)
+```
+
+## getPlayerInfo(playerID=0):
```
akeaswaran, line 83: Do you have to provide defaults for these? I assume any integer is potentially a valid ID, so it may be best to just set this default to `None` (or just not provide a default) and do a check for `None` at the top of the method itself. (This also goes for all of the other similar methods -- I wouldn't provide defaults for anything unless the default is obvious: "isActive"="y", "season"=2022, etc)

examples/test_mlbam.py
```diff
@@ -0,0 +1,404 @@
...
+
+Example:
+
+import sportsdataverse as sdv
+df = sdv.mlb.getAllTimeRoster(teamID=113,startSeason=2020,endSeason=2021)
+print(df)
+"""
+print('getAllTimeRoster')
+#print(sdv.mlb.getAllTimeRoster(teamID=113,startSeason=2020,endSeason=2021))
+df = sdv.mlb.getAllTimeRoster(teamID=113,startSeason=2020,endSeason=2021)
```
akeaswaran, line 402: These tests just print the output so someone can check it manually -- I realize I didn't do this for the CFB stuff, but it may be prudent for us to start writing more robust tests as we look for more people to use this stuff

sportsdataverse/mlb/mlb_loaders.py
```diff
@@ -0,0 +1,47 @@
...
+	saveFile (boolean) = False
+		If saveFile is set to True, the copyright file generated is saved.
+
+	returnFile (boolean) = False
+		If returnFile is set to True, the copyright file is returned.
+	"""
+	url = "http://gdx.mlb.com/components/copyright.txt"
+	resp = download(url=url)
+
+	l_string = str(resp, 'UTF-8')
```
akeaswaran, line 23: It's possible this cast might fail if `resp` is None. Could move this after the None check to avoid that issue

examples/test_mlbam.py
armstjc: These tests were just to ensure that the functions worked as advertised with the MLBAM API.

sportsdataverse/mlb/mlb_loaders.py
armstjc: image
The code right below this line handles that.

docs/docs/mlb/index.md
armstjc: Will fix.

sportsdataverse/mlb/mlbam_reports.py
```diff
@@ -143,7 +107,7 @@ def getBroadcastInfo(season=0,home_away="e",startDate=0,endDate=0):
 		Additionally, endDate must be in YYYYMMDD format. If it is not in that format,
 		the function may not work properly.
 	'''
-	pullCopyrightInfo()
+	#pullCopyrightInfo()
 	#p_df = pd.DataFrame()
 	main_df = pd.DataFrame()
```
akeaswaran, line 112: Outside of the scope of the review here, but:
```python
sd = int(startDate)
ed = int(endDate)
if len(str(sd)) == 8:
```
Why are you converting this to an int and then back to a string? Why not take it as a string param to begin with OR a python Datetime/Date and do your validation for YYYYMMDD that way?

sportsdataverse/mlb/mlbam_stats.py
```diff
@@ -379,158 +340,160 @@ def getCareerPitchingStats(playerID=0,gameType="R"):
...
+# 		else:
+# 			print(f'No results found for the provided playerID. \nTry a diffrient search for better results.')
 		
-		return main_df
+# 		return main_df
 
-def getProjectedHittingStats(playerID=0,gameType="R"):
-	'''
-	Retrives the projected hitting stats for an MLB player, given a proper MLBAM ID
+# def getProjectedHittingStats(playerID=0,gameType="R"):
```
akeaswaran, line 426: Why is all of this stuff commented out?

armstjc: MLBAM API no longer outputs projected stats (this has since been moved to statcast).

test.py
```diff
@@ -1,3 +1,404 @@
 import sportsdataverse as sdv
-cfb_df = sdv.cfb.get_cfb_teams()
-print(cfb_df)
\ No newline at end of file
+
+"""
```
akeaswaran, line 3: This can probably be removed as it seems like it's mainly used for local testing

armstjc: Will fix.

sportsdataverse/mlb/mlbam_stats.py
akeaswaran: I'd just remove all of the code then

sportsdataverse/mlb/mlb_loaders.py
akeaswaran: Right, I understand that -- but if resp is `None`, this method is going to fail at the declaration of `l_string`. See example:

image

armstjc: Ok. I see what you mean.

examples/test_mlbam.py
akeaswaran: Right, but I'm saying it might be good to do some deeper validations with the data that's coming in from MLBAM and make sure what we're promising in the docs is what we're actually returning with each method. EX: actually make sure that the `team_id` column for the dataframe returned here is actually 113, etc.

sportsdataverse/mlb/mlbam_reports.py
armstjc: Issue fixed by repeating the same date technique seen in the NHL part of SDV-py.

sportsdataverse/mlb/mlb_loaders.py
```diff
@@ -19,29 +19,31 @@ def pullCopyrightInfo(saveFile=False,returnFile=False):
...
-			file.writelines(str(l_string))
-
-		with open("mlbam_copyright.txt", "r" ,encoding = "utf-8") as file:
-			mlbam = file.read()
-
-		if saveFile == False:
-			if os.path.exists("mlbam_copyright.txt"):
-				os.remove("mlbam_copyright.txt")
+	try:
+		l_string = str(resp, 'UTF-8')
```
akeaswaran, line 23: The try/catch model here works, but I would really just move the `l_string` declaration under `if resp is not None:`. You're not using `l_string` in the else case where `resp` is None so there's no need to declare it outside of that context

sportsdataverse/mlb/mlbam_reports.py
```diff
@@ -36,20 +36,26 @@ def getTransactionsInRange(startDate=0,endDate=0):
...
-		searchURL = searchURL + f'start_date=\'{ed}\''
-
-		if (endDate - startDate) > 90:
+		#sd = int(startDate)
+		#ed = int(endDate)
+		sd_date = datetime.strptime(startDate, '%m/%d/%Y')
+		ed_date = datetime.strptime(endDate, '%m/%d/%Y')
+		sd = sd_date.strftime("%Y%m%d")
+		ed = ed_date.strftime("%Y%m%d")
+		diff_days = ed_date.date() - sd_date.date()
```
akeaswaran, line 45: also gonna wanna validate that endDate > startDate

sportsdataverse/mlb/mlbam_reports.py
```diff
@@ -36,20 +36,26 @@ def getTransactionsInRange(startDate=0,endDate=0):
...
+		if (diff_days.days)> 30:
 			print('Getting transaction data. This will take some time.')
 		else:
 			print('Getting transaction data.')
+		#print(sd,ed)
+		searchURL = searchURL + f'start_date=\'{sd}\''
+		searchURL = searchURL + f'start_date=\'{ed}\''
+		
+		
 	except:
```
akeaswaran, line 55: You should `return` here and bail out of the method if the inputs are invalid

sportsdataverse/mlb/mlbam_stats.py
```diff
@@ -146,7 +146,7 @@ def getSeasonPitchingStats(playerID=0,gameType="R",season=0):
 		searchURL = searchURL  + f'game_type=\'{gameType}\'&'
 
 	now = datetime.now()
-	if season < 1860:
+	if season < 1860 or season == None:
```
akeaswaran, line 149: also gonna want to have a check for future years

armstjc: What exactly do you mean by "future" years?

akeaswaran (approved):

PR #37, opened 2022-08-02
Armstjc/fix mlb function names - Release Candidate 2

armstjc:


examples/test_mlbam.py
```diff
@@ -285,14 +285,14 @@ def getCareerHittingStats(playerID=0,gameType="R"):
...
 
 import sportsdataverse as sdv
-df =sdv.mlb.getTransactionsInRange(startDate=20200901,endDate=20200914)
+df =sdv.mlb.load_mlbam_transactions(startDate=20200901,endDate=20200914)
 print(df)
 """
-print('getTransactionsInRange')
-print(sdv.mlb.getTransactionsInRange(startDate=20200901,endDate=20200914))
+print('load_mlbam_transactions')
+print(sdv.mlb.load_mlbam_transactions(startDate="09/01/2020",endDate="09/01/2020"))
```
akeaswaran, line 292: Providing two different date formats here

armstjc: That was a fix to reflect code changes that happened when this pull request was happening: https://github.com/sportsdataverse/sportsdataverse-py/pull/33

akeaswaran (approved):
:seemslegit:

saiemgilani:
why do these have loads_? shouldn't it be like mlbam_transactions() ?

armstjc:
> why do these have loads_? shouldn't it be like mlbam_transactions() ?

Because the function/API call requires a range of dates for the call to work, and there isn't another function like this in sdv-py.

saiemgilani:
> > why do these have loads_? shouldn't it be like mlbam_transactions() ?
> 
> Because the function/API call requires a range of dates for the call to work, and there isn't another function like this in sdv-py.

Fine but I reserve the right to change the naming convention here if I feel like it is necessary

saiemgilani (commented):
Fine but reserve the right to change

PR #56, opened 2022-11-15
espn_mbb_game_rosters

saiemgilani:


sportsdataverse/mbb/mbb_game_rosters.py
```diff
@@ -0,0 +1,179 @@
...
+        'first_name', 'last_name', 'full_name', 'athlete_display_name',
+        'short_name', 'weight', 'display_weight', 'height', 'display_height',
+        'age', 'date_of_birth', 'slug', 'jersey', 'linked', 'active',
+        'alternate_ids_sdr', 'birth_place_city', 'birth_place_state',
+        'birth_place_country', 'headshot_href', 'headshot_alt',
+        'experience_years', 'experience_display_value',
+        'experience_abbreviation', 'status_id', 'status_name', 'status_type',
+        'status_abbreviation', 'hand_type', 'hand_abbreviation',
+        'hand_display_value', 'draft_display_text', 'draft_round', 'draft_year',
+        'draft_selection', 'player_id', 'starter', 'valid', 'did_not_play',
```
akeaswaran, line 28: weird whitespacing here

sportsdataverse/mbb/mbb_game_rosters.py
```diff
@@ -0,0 +1,179 @@
...
+        teams_df.loc[row, 'logo_dark_href'] = team[1]['href']
+    teams_df = teams_df.drop(['logos'], axis = 1)
+    teams_df['team_id'] = teams_df['team_id'].astype(int)
+    return teams_df
+
+def helper_mbb_roster_items(items, summary_url):
+    team_ids = list(items['team_id'])
+    team_rosters = pd.DataFrame()
+    for tm in team_ids:
+        team_roster_url = "{x}/{t}/roster".format(x = summary_url,t = tm)
```
akeaswaran, line 128: this all feels like a lot of code to just run under the hood and churn through without some user feedback -- might be a good idea to log every time there's a web request to show intermediate progress?

sportsdataverse/mbb/mbb_game_rosters.py
```diff
@@ -0,0 +1,179 @@
...
+        team_roster_resp = download(team_roster_url)
+        team_roster = pd.json_normalize(json.loads(team_roster_resp).get('entries',[]), sep = '_')
+        team_roster.columns = [col.replace("$ref", "href") for col in team_roster.columns]
+        team_roster.columns = [underscore(c) for c in team_roster.columns.tolist()]
+        team_roster['team_id'] = tm
+        team_rosters = pd.concat([team_rosters, team_roster], axis = 0, ignore_index = True)
```
akeaswaran, line 134: would name these more different things to avoid confusion

sportsdataverse/mbb/mbb_game_rosters.py
```diff
@@ -0,0 +1,179 @@
...
+        'againstTheSpreadRecords',
+        'franchise',
+        'events',
+        'college'
+    ]
+    teams_df = pd.DataFrame()
+    for x in items['team_href']:
+        team = json.loads(download(x))
+        for k in pop_cols:
+            team.pop('{}'.format(k), None)
```
akeaswaran, line 92: can you not just pass k directly here

sportsdataverse/mbb/mbb_game_rosters.py
```diff
@@ -0,0 +1,179 @@
...
+
+    items.columns = [underscore(c) for c in items.columns.tolist()]
+    items = items.rename(
+        columns = {
+            "id": "team_id",
+            "uid": "team_uid",
+            "statistics_href": "team_statistics_href"
+        }
+      )
+    items['team_id'] = items['team_id'].astype(int)
```
akeaswaran, line 67: Reminder: need to ensure this type is maintained throughout the code

sportsdataverse/mbb/mbb_game_rosters.py
saiemgilani: (I'm just assuming you mean about churning through the athlete ID's. It really is like 33 calls per function call. I think the issue is, it doesn't even give you much on the player name just first letter and last name. If it had more details on the players in the roster page, I wouldn't do the cycle.

sportsdataverse/mbb/mbb_game_rosters.py
saiemgilani: I think I had some concerns about it with the $ref but I changed it

sportsdataverse/mbb/mbb_game_rosters.py
saiemgilani: renamed to game_rosters and game_athletes

sportsdataverse/mbb/mbb_game_rosters.py
saiemgilani: fair enough

sportsdataverse/mbb/mbb_game_rosters.py
saiemgilani: I know we're both uncomfortable with one function making that many calls under the hood for the espn_mbb_game_rosters function but I kinda need it to make the rest of the loaders that'll subsidize that to work. then i'll rewrite to use the loader first before making any additional api calls necessary

PR #58, opened 2023-02-17
Check for URL and return error if invalid

### LINKED ISSUE #13: 404 error in games
sportsdataverse.cfb.load_cfb_schedule gives a 404 when trying to get games from 2022, should there be a check and then this just returns nothing?

JeffHCross:
Resolves #13 
Checks for invalid URLs in `sportsdataverse.cfb.load_cfb_schedule` and `sportsdataverse.cfb.load_cfb_pbp`, raising SeasonNotFoundError if the URL is not valid

### DIFF
--- a/sportsdataverse/cfb/cfb_loaders.py
+++ b/sportsdataverse/cfb/cfb_loaders.py
@@ -5,7 +5,7 @@ from tqdm import tqdm
 from typing import List, Callable, Iterator, Union, Optional
 from sportsdataverse.config import CFB_BASE_URL, CFB_ROSTER_URL, CFB_TEAM_LOGO_URL, CFB_TEAM_SCHEDULE_URL, CFB_TEAM_INFO_URL
 from sportsdataverse.errors import SeasonNotFoundError
-from sportsdataverse.dl_utils import download
+from sportsdataverse.dl_utils import download, valid_url
 
 def load_cfb_pbp(seasons: List[int]) -> pd.DataFrame:
     """Load college football play by play data going back to 2003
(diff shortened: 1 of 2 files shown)


JeffHCross:
I didn't notice how `sportsdataverse.cfb.load_cfb_team_info` handles Exceptions until after I had written valid_url, so hopefully you don't mind the different approach.

saiemgilani:
Thank you for this, and I'm sorry it sat for so long — three years is far too long to leave a correct bug report unanswered.

Your diagnosis was right. `load_cfb_pbp([1999])` really did blow up with a raw parquet/HTTP error when a season had no published asset, and guarding the URL before reading was a reasonable fix.

It is fixed on `main` now, though it arrived by a different route than this PR:

- `3173503f` added `_codegen_runtime._read_release_parquet`, which returns `None` when a release asset is missing or 404s, and the generated loaders now collect the missing seasons and emit a `cli_warn("... no data for season(s) [...] (skipped)")`.
- `a75f5031` reworked `download()`'s retry logic so it never retries a definitive 404.

One difference worth naming, because it's a real design choice and not just an implementation detail: `main` **warns and skips** a missing season, where this PR **raised** `SeasonNotFoundError`. Yours is arguably the stricter and more defensible behaviour for a library. If you still want raise-on-missing, it would now be a deliberate change to `releases.yaml` / `_codegen_runtime` rather than a re-open here — `cfb_loaders.py` is generated code these days and carries a `# GENERATED ... DO NOT EDIT` header.

Closing as fixed. Genuinely grateful you took the time to report and patch it.

PR #82, opened 2024-03-11
adds NFL PBP Testing.py

dataengineergaurav:
Code reference taken for this PR: cfb_pbp_testing.py

- Adds nfl_pbp_testing.py

## Summary by CodeRabbit

- **New Features**
	- Introduced a feature for processing and displaying NFL play-by-play data, focusing on special teams plays such as field goals, kickoffs, and punts.

### DIFF
--- a/examples/nfl_pbp_testing.py
+++ b/examples/nfl_pbp_testing.py
@@ -1,0 +1,43 @@
+import pandas as pd
+
+import sportsdataverse as sdv
+
+pd.set_option("display.max_columns", None)
+pd.set_option("display.max_rows", None)
+
+
+def main():
+    processor = sdv.nfl.NFLPlayProcess(gameId=401220403)
+    pbp_init = processor.espn_nfl_pbp()
+    pbp_fin = processor.run_processing_pipeline()
+    # print(pbp_fin)
+    plays_df = pd.DataFrame(pbp_fin["plays"])
+    # print(plays_df)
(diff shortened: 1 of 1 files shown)


saiemgilani:
Thank you for this, and apologies for the long silence.

I'm going to close this one, but I want to be clear that it pointed at something real.

The script itself doesn't quite fit the repo as it stands: it lives under `examples/`, which `pytest.ini` excludes via `testpaths = tests`, so it was never actually collected as a test; and it prints `special_teams.columns` rather than asserting anything, so it couldn't fail. The file it was modelled on, `examples/cfb_pbp_testing.py`, was itself removed in `3173503f`, and `examples/` is notebooks-only now.

But the gap you were poking at is genuine and still open. `tests/nfl/test_nfl_pbp.py` already runs the very same game (401220403) through `NFLPlayProcess.run_processing_pipeline()`, yet only asserts box-score, havoc, and spread — it never checks the special-teams player-name columns. The only test in the repo that touches `fg_block_player_name` / `kickoff_return_player_name` / `punt_block_player_name` is `tests/cfb/test_cfb_player_cols.py`, on the CFB side. So NFL has a real coverage hole that CFB doesn't.

The equivalent contribution today would be a `tests/nfl/test_nfl_player_cols.py` mirroring the CFB one and asserting those columns against the existing fixture — small, and worth doing on its own merits. I've noted it as a follow-up.

Closing this PR, but thank you for spotting it.

PR #86, opened 2025-01-31
change schedule scrape helper to include `offdays`

JackLich10:
## Summary by CodeRabbit

- **New Features**
	- Enhanced ESPN women's basketball calendar retrieval to support both on-days and off-days
	- Improved data collection by fetching from multiple endpoints
	- Ensured unique calendar entries by removing duplicate URLs

### DIFF
--- a/sportsdataverse/wbb/wbb_schedule.py
+++ b/sportsdataverse/wbb/wbb_schedule.py
@@ -138,12 +138,12 @@ def __extract_home_away(event, arg1, arg2):
     return event
 
 
-def espn_wbb_calendar(season=None, ondays=None, return_as_pandas=False, **kwargs) -> pl.DataFrame:
+def espn_wbb_calendar(season=None, onoffdays=None, return_as_pandas=False, **kwargs) -> pl.DataFrame:
     """espn_wbb_calendar - look up the women's college basketball calendar for a given season
 
     Args:
         season (int): Used to define different seasons. 2002 is the earliest available season.
-        ondays (boolean): Used to return dates for calendar ondays
+        onoffdays (boolean): Used to return dates for calendar ondays and offdays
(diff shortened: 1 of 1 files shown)


saiemgilani:
I swear I saw this when you first posted it Jack. By the time I got around to fixing it, I forgot and used a different workaround/pushed it to main. This addressed the MBB/WBB data issues we've been having about not pulling all the games, but may be less than ideal too. I may want to try to use this instead once I compare the differences.

saiemgilani:
Thanks for this, and sorry for the long delay in giving you a proper answer.

Closing it, but with a genuine caveat: the problem you reported is only *partly* solved, and the half you fixed here is the half that never landed.

What did land was a different fix on a different axis — `45c7d5bb` appends `?groups=50` to the `calendar/ondays` URL, which constrains the response to Division I rather than widening the set of dates queried. That addresses "why am I not seeing all the games" from the group-filter side. Your PR addressed it from the date side, by unioning `ondays` with `/calendar/offdays`. Those are orthogonal, so if there are still games sitting on dates that `ondays` alone doesn't return, that gap is real and open today.

Two things kept it from merging as written, both mechanical rather than conceptual: it renames the public parameter `ondays` → `onoffdays`, which is breaking and would leave WBB inconsistent with the six other leagues that all still take `ondays`; and it's WBB-only, while the symptom you described covered MBB as well.

If you're still around and interested, the shape I'd take happily is an additive `offdays=` keyword alongside the existing `ondays=`, applied to both `mbb_schedule.py` and `wbb_schedule.py`. And if you can tell me how many extra dates (and games) `offdays` still surfaces for, say, 2024 and 2025 now that `groups=50` is in place, that measurement alone would be useful — it decides whether this is worth doing at all.

Either way, thank you for the report and the patch.

PR #99, opened 2026-06-12
feat(nfl): cache + auto-renew api.nfl.com token; NFL_ACCESS_TOKEN env override

saiemgilani:
## Summary

Makes `api.nfl.com` auth **automatic and under-the-hood**, and adds an opt-in env override — while confirming Fox / HockeyTech need nothing (static keys).

### NFL.com token caching (the real win)
Previously every `nfl_*` / `nfl_api_*` call POSTed to `/identity/v3/token` to mint a fresh bearer token before the data call. Now:
- The token is **minted once and cached in-process**, then **auto-renewed ~2 min before its JWT `exp`** (read from the token itself, no signature check). Back-to-back calls reuse one token. No setup, no manual refresh.
- Thread-safe (module lock); cache is keyed by the resolving client key so swapping credentials never serves a stale token.
- `nfl_clear_token_cache()` and `nfl_token_gen(force_refresh=True)` for explicit control.

### Env overrides (all optional)
- **`NFL_ACCESS_TOKEN`** *(new)* — inject a pre-minted bearer token verbatim (skips mint + cache; you manage its lifetime). Highest precedence; ignored when explicit `client_key`/`client_secret` args are passed.
- `NFL_CLIENT_KEY` / `NFL_CLIENT_SECRET` *(existing)* — mint with your own credentials instead of the bundled public `WEB_DESKTOP` pair.

### Fox key single-sourced
`cfb_fox_ext.FOX_DATA_KEY` now imports from `_fox_layout.DATA_KEY` — the bundled public Fox key + its `SDV_PY_FOX_DATA_KEY` env override live in one place instead of two. **Fox and HockeyTech are static-key APIs** (a `?apikey=` query param, no token, no expiry), so there's nothing to cache/renew — they already work with zero interaction out of the box, and the env override is just an escape hatch for upstream key rotation.

## Test plan
- [x] `tests/nfl/test_nfl_token_cache.py` (offline, mocks the mint): caching, `force_refresh`, exp-driven renewal, credential-keyed invalidation, `NFL_ACCESS_TOKEN` short-circuit + precedence, clear-cache, headers reuse, `_jwt_exp` parsing — 10 passed.
- [x] NFL + CFB offline suites: 207 passed, 58 skipped.
- [x] `ruff` clean; Fox dedup verified (`cfb_fox_ext.FOX_DATA_KEY is _fox_layout.DATA_KEY`).

> Independent of #98 (NFL native cutover) — different files, no conflict. The only overlap is the `0.0.58 (in development)` CHANGELOG heading both branches add; trivial additive merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Implemented automatic in-process caching and auto-renewal of NFL API bearer tokens with optional environment variable override.
  * Added token cache management helpers for explicit refresh and clearing operations.

* **Documentation**
  * Updated changelog for upcoming release documenting NFL token caching functionality and internal data consolidation improvements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 **Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
 
+- [0.0.58 (in development)](#0058-in-development)
+  - [NFL — automatic `api.nfl.com` token caching + `NFL_ACCESS_TOKEN` override](#nfl--automatic-apinflcom-token-caching--nfl_access_token-override)
+  - [Internal — Fox data key single-sourced](#internal--fox-data-key-single-sourced)
 - [0.0.57 Release: June 10, 2026](#0057-release-june-10-2026)
   - [Fox Sports Bifrost wrappers (CFB, NBA, MBB, NHL, MLB)](#fox-sports-bifrost-wrappers-cfb-nba-mbb-nhl-mlb)
     - [CFB — Fox as a backup source for the EPA/WPA play processor (`fox_cfb_play_process`)](#cfb--fox-as-a-backup-source-for-the-epawpa-play-processor-fox_cfb_play_process)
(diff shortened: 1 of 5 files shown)


saiemgilani:
Folded into #98 (commit a228958) so the NFL.com native cutover + its token-caching/auth ergonomics ship together. Closing in favor of #98.

PR #123, opened 2026-06-22
refactor(nfl): drop the enrich_nfl_pbp method="snapshot" vestige

saiemgilani:
Removes the dead `method="snapshot"` branch from `enrich_nfl_pbp` — an early design-fork value never implemented (the default `"lead_diff"` already achieves nflfastR parity).

- Dedicated `NotImplementedError` branch + docstring/Raises mentions removed.
- Any non-`lead_diff` value (incl. `snapshot`) now raises a plain `ValueError`.
- `method=` param retained for forward extensibility.
- Test updated (`snapshot` -> `ValueError`).

ruff + mypy clean; enrich tests 15 passed; codegen `--check` clean.

Not included: 2a (ESPN-path xYAC) — blocked by the deferred `air_yards` gap (needs FTN-charting air_yards, 2022+), a separate decision.

## Summary by CodeRabbit

* **Refactor**
  * Simplified the `enrich_nfl_pbp()` method validation. The API now supports only `method="lead_diff"`; previously supported methods are no longer accepted and will raise an error.

* **Tests**
  * Updated test suite to reflect new validation behavior.

### DIFF
--- a/sportsdataverse/nfl/ep_wp.py
+++ b/sportsdataverse/nfl/ep_wp.py
@@ -2447,10 +2447,9 @@ def enrich_nfl_pbp(
             WP inputs ``score_differential`` / ``game_seconds_remaining`` /
             ``spread_line`` / ``receive_2h_ko``).
         method: ``"lead_diff"`` (default) — the native, nflfastR-faithful
-            EPA/WPA derivation and the canonical parity path; it is the only
-            implemented method.  ``"snapshot"`` is accepted for API stability
-            but is **not implemented** (it has no nflfastR-parity reference) and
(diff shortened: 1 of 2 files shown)


saiemgilani:
Re: the two CodeRabbit suggestions to restore `method="snapshot" -> NotImplementedError` — **declining intentionally.** This PR's whole purpose is to remove the `snapshot` vestige: it was an abandoned early design-fork value with no nflfastR-parity reference and no implementation path, not a "reserved/coming-later" feature. Treating it as a plain unsupported value (`ValueError`, like any other non-`lead_diff` method) is the intended contract here. The `method=` param is retained for genuine future methods. Done at the maintainer's direction.

PR #145, opened 2026-06-29
feat(validation): domain ③ Phase 1.5 — finer R-lint root, cumulative heuristic, lint targets, nfl_model_pbp parity

saiemgilani:
## Validation Domain ③ — Phase 1.5

Closes out domain ③ with four mostly-independent improvements, stacked on #144 (Phase 3). Each rides the existing `Finding`/CLI/registry/oracle machinery — no new toolchain.

> **Stacked PR.** Base = `feat/validation-domain3-phase3` (#144 ← #143 ← #142 ← #141). Review/merge the parents first; the diff here is Phase 1.5 only.

### A — Finer R-lint statement root
`leakage_r._root` now also stops the parent-walk at brace-block boundaries (`block_ids` = parents of `'{'` tokens), so a grouped pipe chain no longer **masks** an ungrouped `lag`/`cum*` elsewhere in the same `function(){...}` — the low-recall case the Phase 3 whole-branch review flagged as material on real (function-wrapped) R modeling code. Validated against a real R 4.6.0 `getParseData` capture: `fnwrap.R` → 1 (the ungrouped lag inside the function, previously 0/masked), and **backward-compatible** `leaky.R`→2 / `clean.R`→0 (top-level fixtures have no braces). The docstring documents the new granularity and a rare residual false-positive (inline `{}`/lambda inside a grouped pipe → WARN, dismissed by Tier-2).

### B — Cumulative-column boundary heuristic
`boundary_leakage.run` now inspects `ctx.cumulative_columns` (previously "accepted but not yet inspected"): a group whose first value exceeds the prior group's last value (a non-reset — possible carried accumulation across a game boundary) → **WARN `needs_judgment`** (cross-group only; reset semantics are column-specific and heuristic). Registers `cfb_model_pbp.cumulative_columns=("game_play_number",)` (resets to 1 each game). Fixtures + tests cover reset→clean, non-reset→1, and multi-boundary→`metric==2.0` (one finding per column, not per boundary).

### C — Real Python `LINT_TARGETS`
Registers the two real Python lint surfaces so the linters actually run on real code: `nfl_native_pbp` (env-rooted `${SDV_VALIDATION_NFL_DATA_ROOT}/python/native_pbp`) and `sdv_nfl_ep_wp` (repo-relative `sportsdataverse/nfl/ep_wp.py`). Both dispatch through the existing `cli lint --target`.

### D — `nfl_model_pbp` dataset + wired `NflfastrOracle` (real ep/epa/wp parity)
Registers `nfl_model_pbp` and **wires the previously-degrading `NflfastrOracle`** to the local nflverse reference, so `numeric_parity` is a real ep/epa/wp/vegas_wp/cp comparison vs nflfastR — not just schema/boundary. The oracle's `reference_frame` now:
- expands env vars and **column-prunes** the 372-col all-history reference via a lazy scan (only join keys + `column_map` columns are materialized);
- **aligns join-key dtypes** (the reference's Float64 `play_id` → the producer's Int64 — R serializes ints as `123.0`; without this the `game_id`+`play_id` join would silently mismatch → a *false* parity-pass). The cast is reference→producer and unconditional-on-mismatch (a non-coercible type raises loudly rather than yielding a vacuous join);
- **degrades to `None` → INFO** when the reference glob matches no files (missing cache / unset env var) instead of crashing `run_dataset`.

The parity join was verified non-vacuous: **94,976 of 94,978 producer rows** match the reference. Sub-floor correlation matches are **expected** WARN `needs_judgment` (the producer is model-domain feature-substituted while the reference is raw all-play-type pbp — documented on `thresholds`; recalibration belongs with the Tier-2 parity-reviewer). A gated live parity test runs the real comparison where the data is present (skips on bare CI); synthetic monkeypatched unit tests cover the dtype-cast / empty-glob / no-shared-keys paths in CI with no data.

### Config
One env var, `SDV_VALIDATION_NFL_DATA_ROOT` = the nfl-data repo root, serves the dataset glob (`out/model_pbp_*.parquet`), the `native_pbp` lint target, and the oracle reference (`python/.cache/nflverse_pbp/pbp_*.parquet`). sdv-py's own `ep_wp.py` is repo-relative.

### Quality
Built via subagent-driven development (per-task spec + code-quality review, plus a final whole-branch review). The riskiest pieces were validated empirically before coding (finer-root capture) and adversarially after (the 95k-row join proven non-vacuous; range constraints checked against real data: ep [-3.4,6.5]⊂[-10,10], epa [-12.8,8.8]⊂[-15,15]). Gate: **67 contract tests / 1 gated-skip, mypy 30 files, ruff check + format**, `uv.lock` untouched.

### Follow-ups (not here)
- Tier-2 `leakage-reviewer` / `parity-divergence` judgment agents that consume these WARN `needs_judgment` findings; recalibrate the NFL parity floors (or model-domain-filter the reference) there.
- Optional: restrict the R-lint `block_ids` to function-definition braces only (drop the rare inline-`{}` false positive).
- Register the cfb-data R `LINT_TARGET`; broaden the nfl glob to backfill seasons.
- Pre-existing: `numeric_parity` treats a constant-column `nan` corr as passing (a latent blind spot, unrelated to this PR).

## Summary by Sourcery

Extend validation coverage for NFL/CFB datasets and R/Python linting by tightening leakage detection heuristics and wiring real data/parity surfaces.

New Features:
- Add cumulative boundary leakage detection on declared cumulative columns with judgment-required WARN findings.
- Register real Python lint targets for nfl-native play-by-play and sdv-py ep/wp code.
- Register the nfl_model_pbp dataset with schema, range constraints, and numeric parity checks against an nflfastR-based oracle.
- Gate a live nfl_model_pbp parity test that runs when real NFL data is available.

Enhancements:
- Refine the R leakage lint to use per-brace-block statement roots, improving detection of ungrouped lag/cumulative calls inside functions while documenting rare residual edge cases.
- Improve the NFL parity oracle to lazily load and column-prune reference data, align join-key dtypes to the producer, and degrade safely when no reference files or shared keys are present.
- Document parity thresholds and expected sub-floor correlations for nfl_model_pbp vs the raw nflverse reference.
- Broaden boundary leakage validation to handle frames with both lag and cumulative columns and update tests to cover new behaviors.

Tests:
- Add contract tests and fixtures for cumulative boundary leakage behavior, the refined R leakage lint rooting inside function-wrapped code, nfl oracle reference-frame casting/empty-reference behaviors, registry wiring of lint targets and nfl_model_pbp, and a gated live nfl_model_pbp parity run.


saiemgilani:
Superseded by #149, which re-linearizes this stack's work (with all bot-review fixes) onto main — the divergent stacked-squash dropped content, so the chain was consolidated. Closing as landed via #149.

PR #259, opened 2026-07-13
fix(mbb): pace NcaaFetcher proxy rotation with an inter-rotation backoff

saiemgilani:
## Summary

`NcaaFetcher._get_with_rotation` rotated to the next proxy on any non-clean
response or transport exception with zero delay. During a ban wave this
fires `len(pool) + max_retries` requests back-to-back, worsening the ban and
burning the whole ~50-proxy pool instantly before hard-stopping.

## Fix

- New `NcaaFetchConfig.rotation_backoff: float = 1.0` (seconds), overridable
  via `SDV_PY_NCAA_ROTATION_BACKOFF`.
- `_get_with_rotation` sleeps `rotation_backoff` seconds before every
  proxy-rotation attempt except the first, so a clean first hit stays
  instant (no added latency on the happy path) and only the retry path is
  paced.

## Tests

- Existing `test_ban_marker_response_rotates_proxy` /
  `test_all_proxies_exhausted_raises` updated to pass `rotation_backoff=0.0`
  so they stay instant.
- New: backoff sleeps exactly twice (once per retry) on a
  ban-ban-clean sequence, verified via a monkeypatched `time.sleep`
  recorder.
- New: no sleep at all on a clean first hit.
- New: `SDV_PY_NCAA_ROTATION_BACKOFF` env override (valid + invalid value).

`uv run pytest tests/mbb/test_mbb_ncaa_fetch.py -q` — 25 passed in ~2s (fast,
no real sleeps). `ruff check` and `mypy` clean.

## Known follow-up

`uv run python tools/codegen/generate.py --check` flags stale
`docs/docs/{mbb,wbb}/reference/additional.md` — both embed
`NcaaFetchConfig`'s generated constructor signature, which now includes
`rotation_backoff`. Docs were intentionally left unregenerated in this PR
pending review; run `uv run python tools/codegen/generate.py` and commit
the diff before merge.

## Test plan

- [ ] `uv run pytest tests/mbb/test_mbb_ncaa_fetch.py -q`
- [ ] `uv run ruff check sportsdataverse/mbb/mbb_ncaa_fetch.py tests/mbb/test_mbb_ncaa_fetch.py`
- [ ] `uv run mypy sportsdataverse/mbb/mbb_ncaa_fetch.py`
- [ ] Regenerate docs (`uv run python tools/codegen/generate.py`) and commit before merge

## Summary by CodeRabbit

* **New Features**
  * Added `rotation_backoff` to control pacing between proxy-rotation retry attempts.
  * Added environment-variable support to configure `rotation_backoff` (with safe fallback to the default).
* **Bug Fixes**
  * Added retry delays only when needed, avoiding unnecessary pauses after a successful first request.
* **Tests**
  * Expanded proxy-rotation coverage, including assertions for sleep behavior and environment-based overrides.
* **Documentation**
  * Documented the new `rotation_backoff` parameter in the NCAA/WBB `NcaaFetchConfig` references.

### DIFF
--- a/docs/docs/mbb/reference/additional.md
+++ b/docs/docs/mbb/reference/additional.md
@@ -971,7 +971,7 @@ The set of players on the floor, as an opaque id string
 
 Game location (`Game.LocationType`, `Game.scala:36-38`).
 
-### `NcaaFetchConfig(cache_dir: 'Optional[Path]' = None, proxy_url: 'Optional[str]' = None, proxybonanza_key: 'Optional[str]' = None, proxybonanza_pkg: 'Optional[str]' = None, timeout: 'int' = 45, impersonate: 'str' = 'chrome', max_retries: 'int' = 2, transport: 'Optional[FetchTransport]' = None) -> None` {#NcaaFetchConfig}
+### `NcaaFetchConfig(cache_dir: 'Optional[Path]' = None, proxy_url: 'Optional[str]' = None, proxybonanza_key: 'Optional[str]' = None, proxybonanza_pkg: 'Optional[str]' = None, timeout: 'int' = 45, impersonate: 'str' = 'chrome', max_retries: 'int' = 2, rotation_backoff: 'float' = 1.0, transport: 'Optional[FetchTransport]' = None) -> None` {#NcaaFetchConfig}
 
 Runtime configuration for the stats.ncaa.org fetch layer.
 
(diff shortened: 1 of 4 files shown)


saiemgilani:
Addressed the CodeRabbit review in 03970bf6: `_from_env()` now rejects a non-finite `SDV_PY_NCAA_ROTATION_BACKOFF` (`inf`/`-inf`/`nan`) via `math.isfinite`, keeping the default rather than letting it reach `time.sleep()` on the retry path (which would raise `OverflowError`/`ValueError`). Extended `test_rotation_backoff_env_override` to cover all three. Fetch suite: 25 passed; ruff + mypy clean.

PR #263, opened 2026-07-16
feat(cfb): add the load_cfb_ratings dataset loader

saiemgilani:
## Summary

Half B of the CFB ratings publish program — the **read** side of the `cfb_ratings` tag that [`cfbfastR-cfb-data#21`](https://github.com/sportsdataverse/cfbfastR-cfb-data/pull/21)'s `cfb_model_publish ratings` builder publishes. Follows #262, which made `cfb_ratings()` work against the released `espn_cfb_pbp` asset in the first place.

- **`releases.yaml`** — `load_cfb_ratings` (base `sdv_releases`, url `cfb_ratings/cfb_ratings_{season}.parquet`, `min_season: 2004` — the `espn_cfb_pbp` floor the ratings are computed from).
- **`loader_schemas.yaml`** — the 15-column returns-schema, **introspected from a real parquet emitted by the builder** rather than derived by hand.
- Regenerated `cfb_loaders.py` + `parsed/cfb.py` + docs.

## One loader, not three

The plan sketched three. It's one:

- **`cfb_adv_efficiency` — dropped as redundant.** The union of `efficiency_ratings` + `special_teams_ratings` + `fei_ratings` is **10 columns**, every one of which is already in `cfb_ratings`' **15** at an identical dtype (the orchestrator's final `.select()` drops none of them). It would have published a strict subset of tag #1.
- **`cfb_recruiting_proj`** waits on its own builder.

## The contract test

The existing codegen suite covers URL shape and that loader modules render valid Python. Nothing asserted that **a loader's declared returns-schema still matches what the producer actually emits** — and the published parquet is written straight from `cfb_ratings()`'s frame, so any divergence is a docs lie rather than a formatting nit.

`test_declared_schema_matches_the_producer_output_schema` pins the two together on column order **and** dtype. Verified it catches both failure modes it exists for: an added producer column, and a `team_id` `Utf8 -> Int64` flip (the ID-dtype footgun in CLAUDE.md).

## Gates

Full `tests/cfb/` + `tests/codegen/`: **611 passed**, 53 skipped, 2 xfailed. `codegen --check` clean. `ruff check` + `format --check` clean. `mypy` clean. `uv.lock` untouched.

## Note for the next person regenerating

`generate.py` renders `parsed/<league>.py` by introspecting the **imported** package, so the run that first writes a new loader into `cfb_loaders.py` still renders `parsed/cfb.py` *without* it — and `--check` then reports `parsed/cfb.py` stale even though `git status` shows it unmodified. A **second `generate.py` pass** converges it. (Confirmed the drift is caused by this change, not pre-existing: `--check` on pristine `origin/main` is clean.)

`load_cfb_ratings` is 404-safe like every release loader, so until the backfill publishes the tag it returns empty-with-a-warning rather than raising.

## Summary by CodeRabbit

- **New Features**
  - Added the `load_cfb_ratings` college football dataset loader.
  - Supports loading ratings by season as Polars or pandas data.
  - Handles unavailable season releases safely with warnings.
  - Added public exports and reference documentation, including output fields and usage examples.

- **Documentation**
  - Updated the CFB documentation index and unreleased changelog.
  - Replaced the previous ratings function reference with the new loader documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 **Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
 
 - [Unreleased](#unreleased)
+  - [CFB — `load_cfb_ratings` dataset loader](#cfb--load_cfb_ratings-dataset-loader)
   - [NBA / WNBA — CTG play context (T3.6): possession/shot/lineup/player tables + start-type oracle](#nba--wnba--ctg-play-context-t36-possessionshotlineupplayer-tables--start-type-oracle)
   - [Fixes](#fixes)
   - [Dependencies](#dependencies)
@@ -175,6 +176,20 @@
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
 ## Unreleased
+
+### CFB — `load_cfb_ratings` dataset loader
+
+- **`load_cfb_ratings(seasons)`** — loads the published `cfb_ratings` release
+  (one row per team per season, 2004–): opponent-adjusted offensive / defensive /
+  special-teams EPA, FEI, `games`, `off_pace`, dense `off_rank` / `def_rank` /
+  `net_rank`, and `net_z`. The tag is produced by `cfbfastR-cfb-data`'s
+  `cfb_model_publish ratings` builder running sdv-py's own
+  :func:`sportsdataverse.cfb.cfb_ratings` over the released `espn_cfb_pbp`
+  play-by-play, so the loader's returns-schema is the compute function's output
+  schema — a contract test pins the two together (both column order and dtype)
+  so a producer change can't silently leave the published returns-table lying.
+  Like every release loader it is 404-safe: seasons with no published asset are
+  skipped with a warning rather than raising.
 
 ### NBA / WNBA — CTG play context (T3.6): possession/shot/lineup/player tables + start-type oracle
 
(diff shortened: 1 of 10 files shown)


saiemgilani:
Triaged both CodeRabbit nitpicks — **declining both**, verified against the code rather than the guideline text they cite.

**1. `pl.Utf8` -> `pl.String` (test_cfb_loaders_ratings.py:29) — declined.** `pl.Utf8` is not a legacy 0.18-era API; it is an **alias of `pl.String`** in the installed polars:

```
polars 1.42.0
pl.Utf8 is pl.String -> True      # same object
# no DeprecationWarning on access
```

CLAUDE.md's polars-1.x migration table does not list this rename, and the repo is `pl.Utf8` **1114** occurrences to `pl.String` **7** — including `cfb_ratings.py`'s own `_RATINGS_OUTPUT_SCHEMA`, which is exactly what this test compares against. Switching one test file would make it inconsistent with the schema it asserts on, for zero behavioural change (CodeRabbit's own note concedes "equality checks will still pass since they are aliased").

**2. Remove `from __future__ import annotations` (line 10) — declined.** It's the dominant local convention: **37 of 52** files under `tests/cfb/` use it, including `test_cfb_ratings.py`, the direct sibling this file pairs with. No repo guideline bans it.

Both were CodeRabbit's own "Trivial / Low value" tier and neither changes behaviour.

---

**On the red `release-manifest audit (informational)`** — that one is real, expected, and self-resolving. It reports:

```
release manifest drift: 1 orphan manifest tag(s) (loader -> dead release):
  cfb_ratings
```

This PR adds the loader; the `cfb_ratings` release does not exist until the 2004-> backfill runs (Half C). The job is `continue-on-error: true` precisely so a loader/release sequencing gap surfaces without red-failing the PR, and it clears the moment the tag is published. The other 39 orphans it lists are pre-existing.

PR #296, opened 2026-07-22
fix(rds): stream write_rds to its sink instead of buffering the whole object

saiemgilani:
## Problem

`write_rds` cannot serialize a season-sized frame. `_RdsWriter` accumulates the entire XDR stream as a list of small `bytes` objects — one per value, each carrying ~33 bytes of interpreter overhead — and `payload()` then joins them into a **second, contiguous copy** of the whole stream before a single write.

Measured on NHL's 2026 play-by-play (**1,100,073 × 94**, 0.69 GB in Arrow), on a 15 GiB box:

```
Out of memory: Killed process (python3)
  total-vm:31471552kB, anon-rss:13774772kB
```

The kill lands *mid-write*, so the season's parquet is on disk and its rds silently isn't — the failure mode that blocked moving the NHL data pipeline onto the Python compiler.

## Fix

`_raw` was already the single funnel for every write, so this is an optional sink on the writer rather than a rewrite:

- **With a sink**, bytes go straight out; **without one**, the old buffering path is untouched and `payload()` still serves the byte-golden fixtures.
- Writes are **batched into 4 MB blocks** first. Serialization emits a few bytes at a time, and handing each one to the gzip compressor individually is ~5 orders of magnitude more calls for no benefit. Memory stays bounded by that constant rather than by the frame.
- **Atomic write.** Streaming introduces a failure mode buffering could not have: a mid-write error would leave a truncated `.rds` where previously no file was created at all. A truncated rds is worse than a missing one — it reads as a corrupt object and can be published as if complete — so the write goes to a temp sibling and renames on success.

## Result

| | before | after |
|---|---|---|
| `write_rds` on NHL pbp | +6.8 GB → **OOM-killed** | **−0.03 GB** |
| write-phase peak | — | 2.60 GB |
| 300k-row throughput | 1.92 s | 2.03 s |

## Verification

- **Byte-identical** to the buffered writer across 26 combinations: frames covering nulls, unicode, empty, dates, wide, and a 300k-row frame that crosses the block boundary repeatedly; compressed and uncompressed; with and without a class vector.
- `readRDS` returns the full `1100073 x 94` frame with `fastRhockey_data,tbl_df,tbl,data.table,data.frame` intact.
- A failed write leaves neither a target file nor a stray temp.
- 104 passed, 3 skipped (`-k "rds or release"`); ruff, ruff format, and mypy clean. `uv.lock` untouched.

## Summary by Sourcery

Stream RDS serialization directly to a file sink with atomic writes to avoid buffering large datasets in memory and prevent truncated output files.

New Features:
- Support writing RDS byte streams to arbitrary binary sinks via a new write-only sink protocol and streaming-capable writer.

Bug Fixes:
- Fix out-of-memory failures when serializing season-sized data frames by eliminating full in-memory buffering of the RDS byte stream.
- Prevent truncated or corrupt RDS outputs by writing to a temporary file and atomically renaming it on successful completion.

Enhancements:
- Batch serialized bytes into fixed-size chunks before writing to the sink to reduce write/compression overhead while keeping memory usage bounded.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when writing RDS outputs by writing to a temporary file first and replacing the final output atomically only on success.
  * Failed writes now automatically clean up temporary partial files to prevent incomplete/corrupted outputs.

* **Performance**
  * Reduced memory usage during RDS serialization by streaming data incrementally rather than buffering the full output in memory.
  * Improved handling of compressed vs. uncompressed output to keep gzip behavior consistent with the final output.

### DIFF
--- a/sportsdataverse/_rds.py
+++ b/sportsdataverse/_rds.py
@@ -24,7 +24,7 @@ from datetime import datetime, timezone
 from functools import partial
 from pathlib import Path
 from collections.abc import Callable, Iterable, Mapping, Sequence
-from typing import Union
+from typing import Protocol, Union
 
 import numpy as np
 import polars as pl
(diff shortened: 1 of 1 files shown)


saiemgilani:
Thanks — addressed in 9423bdce. Taking three, declining two, and one of them caught a bug of mine.

**Accepted**

- **Guard `payload()` on a streaming writer.** Agreed, and worth more than a docstring: silently returning an empty buffer produces an "object" that looks serialized and can be published. It now raises.
- **`flush()` shouldn't depend on callers.** Early return when there's no sink; the `type: ignore` goes away with it.
- **Unique temp name.** Good catch, and not hypothetical in this repo's pipelines — the fixed `.{name}.partial` meant two processes writing the same path (two compiles sharing an `out_dir`) would interleave bytes into one file and both report success. Now `tempfile.mkstemp` in the destination dir.

**Bug your temp-file comment surfaced**

`gzip` stamps the output filename into its header, so opening the temp path directly embedded `.pbp.rds.partial` where the buffered writer embedded `pbp.rds`. My original parity check compared *decompressed* bytes and sailed right past it. `GzipFile` is now constructed with the final name over the temp fd, and the check compares whole compressed files (MTIME masked, since it's inherently volatile):

```
embedded gzip filename     : same_path.rds
full gzip bytes identical  : True
```

**Declined**

- **`BaseException` → `Exception`.** The handler re-raises, so nothing is swallowed; narrowing it would *leak* a temp file on `KeyboardInterrupt`, which is the opposite of the intent. Keeping it broad is deliberate.
- **Configurable `_SINK_BUFFER_BYTES`.** No workload yet suggests tuning, and it's a class attribute, so it's already overridable if one appears. Not adding a knob on speculation.

Gate after the changes: 104 passed / 3 skipped (`-k "rds or release"`), ruff + ruff format + mypy clean, `uv.lock` untouched.

PR #297, opened 2026-07-22
fix(rds): keep the umask-derived file mode when writing via the temp file

saiemgilani:
Follow-up to #296.

## Problem

The atomic-write change silently tightened permissions on every rds it produced. `tempfile.mkstemp` creates `0600`, and `rename` preserves the mode — so output that used to land at the usual umask-derived `0644` (what opening the destination directly gave) became owner-only:

```
-rw------- 1 root root 49537801 play_by_play_2026.rds
```

It survived review because it never fails anything visible: publishing works, since the same user uploads the asset. It only shows up as a mode on disk. Anything reading a compiled tree as another user or service would be denied.

## Fix

`chmod` the temp file to `0666 & ~umask` right after `mkstemp`, matching what a plain `open()` would have produced.

## Verification

Against the pre-#296 writer, same input, both compression modes:

```
compress=True : new=0o644 old=0o644 match=True bytes_identical=True
compress=False: new=0o644 old=0o644 match=True bytes_identical=True
```

104 passed / 3 skipped (`-k "rds or release"`); ruff, ruff format, mypy clean.

## Summary by Sourcery

Bug Fixes:
- Restore umask-derived file permissions when writing RDS files through a temporary file created with mkstemp, avoiding overly restrictive owner-only modes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file permission handling when exporting RDS files to ensure consistent access behavior between the temporary staging file and the final output.
  * Updated atomic export behavior to use a safer temporary-file creation strategy, reducing the chance of unexpected permission-related access issues.

### DIFF
--- a/sportsdataverse/_rds.py
+++ b/sportsdataverse/_rds.py
@@ -20,8 +20,8 @@ from __future__ import annotations
 
 import gzip
 import os
+import secrets
 import struct
-import tempfile
 from contextlib import nullcontext
 from datetime import datetime, timezone
 from functools import partial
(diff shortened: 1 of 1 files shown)


saiemgilani:
Both of you flagged the umask handling and you were right — reworked in ec7851f2.

Two problems with what I had: `os.umask()` mutates **process-global** state, so any other thread creating a file in that window gets the wrong mode; and the read-modify-write *guessed* `0o022` as the restore value instead of the caller's actual umask. @sourcery-ai's `os.umask(0)` form fixes the guess but keeps the race.

Neither is necessary, so I took @coderabbitai's direction and dropped `umask` from the write path entirely. `tempfile.mkstemp` was only ever there for a unique name — and it's what forced `0600` to begin with. Claiming the name directly instead:

```python
fd = os.open(candidate, os.O_CREAT | os.O_EXCL | os.O_WRONLY, 0o666)
```

lets the kernel subtract the caller's umask exactly as `open(path, "wb")` does. No global state touched, no race, no guess. The retry loop preserves `O_EXCL`'s collision guarantee, though a clash requires a 64-bit token to repeat.

Verification — the key one being that the mode now *tracks* the umask rather than landing on a hardcoded `0644`:

```
compress=True : new=0o644 old=0o644 match=True bytes_identical=True
compress=False: new=0o644 old=0o644 match=True bytes_identical=True
umask 077 -> 0o600            (kernel-applied, not hardcoded)
process umask untouched (still 0o22)
no stray .partial files
```

104 passed / 3 skipped (`-k "rds or release"`); ruff, ruff format, mypy clean.

On the "wrap it in a helper" suggestion — done as `_create_temp_sibling()`, which now carries the whole rationale (why not `mkstemp`, why not `umask`) in one place rather than as an inline dance.

PR #300, opened 2026-07-28
fix(nba): three real-data bugs in the raw-store impact path

saiemgilani:
Both surfaced only when validating a real build against real captures; both were silent — no exception, plausible-looking output.

## 1. Season-level captures are keyed by START year

The raw store keys its two halves differently:

| half | key | `…/2024/` holds |
|---|---|---|
| per-game (`playbyplayv3`) | season **END** year (from the game id) | 2023-24 |
| season-level (`leaguegamelog`) | season **START** year (the year passed to the API) | **2024-25** |

`_season_index_from_store` assumed end year for both, so offline discovery returned **the next season's game ids**. They resolve fine — they're just the wrong games, so an entire season compiles against them without an error.

Downstream this made the player↔team join match **0 / 2460**, and `box_features` fell through its `otherwise(0.0)` branch: every player got identical zeroed stats, collapsing SPM to **1 distinct value** and BPM to **30** (one per team).

Fixed, documented on `nba_raw_store_season_frame` (its `season` arg is now stated as START year, with a warning about the split), and pinned by a regression test that ties the halves together — with both neighbouring captures present, discovery must return the matching season *and* the ids must live in the per-game directory for that same season.

## 2. Per-100 features used team slot-minutes as the denominator

A team-game's `min` sums its **five on-court slots** (240 in regulation), not elapsed game minutes. `box_features` divided by it directly, so a wire-to-wire player was credited with a fifth of the possessions he faced — every per-100 rate **5× high**.

SPM never showed it (`train_spm` fits coefficients on these same features, so a uniform scale is absorbed). `nba_bpm` feeds them into **fixed published coefficients**, so it landed whole:

| | before | after |
|---|---|---|
| BPM, qualified (min≥1000) | **-157 .. 184** | **-11.6 .. 17.3** |
| Jokić 2023-24 BPM | 65.4 | 17.3 (**leads**, as in the real race) |

The fixtures hid it by giving a player `min = 240` — impossible in a 48-minute game, and it made `min / team_min` accidentally equal 1. Their own comments stated the correct intent (*"player plays all minutes → player_poss = full team possessions"*), so they now use realistic minutes and **every per-100 assertion in them is unchanged**. Added an oracle: a wire-to-wire player on a 100-possession team-game must get exactly 100 possessions.

## Verification

`700 passed, 41 skipped` in `tests/nba/`; ruff clean. On a real 2023-24 build: join 2460/2460, `pts` 563 distinct (was 1), `spm` 570 / `bpm` 572 distinct, all 572/572 non-null, and the WAR leaderboard reads SGA / Brunson / George / Jokić / Giannis.

Consumers: hoopR-nba-stats-data#21, hoopR-nba-stats-raw#3.

## Summary by Sourcery

Fix NBA raw-store season indexing and correct per-100 possession denominator used in box score features.

Bug Fixes:
- Correct per-100 possession calculations in box_features by basing player possessions on elapsed minutes rather than team slot-minutes, preventing 5x inflation of per-100 rates.
- Fix season index lookup for leaguegamelog data to use the season start-year directory when given an end-year season, ensuring the season compile uses the correct games rather than the neighboring season.

Enhancements:
- Clarify documentation for nba_raw_store_season_frame to spell out the differing season key conventions between per-game and season-level raw-store captures.

Tests:
- Adjust NBA box log tests to use realistic player minutes and add an oracle test that a wire-to-wire player on a 100-possession team-game gets matching possessions and per-100 scoring.
- Add regression tests ensuring season discovery reads from the correct start-year season directory and that returned game IDs align with the per-game store layout.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected player possession and per-100 denominators to reflect five on-court player slots.
  * Fixed NBA raw store season discovery to map end-year requests to the corresponding start-year data.

* **Documentation**
  * Clarified that `season` uses the season start year, with notes about end-year/off-by-one expectations.

* **Tests**
  * Updated fixtures and per-100 assertions for corrected minute and possessions math.
  * Added regression coverage for the season directory/id mapping and added an end-to-end possession oracle scenario.

### DIFF
--- a/docs/docs/nba/reference/additional.md
+++ b/docs/docs/nba/reference/additional.md
@@ -2890,7 +2890,7 @@ same committed tree the per-game compile reads.
 | Parameter | Type | Default | Description |
 |---|---|---|---|
 | `endpoint` | `str` |  | stats.nba.com endpoint slug (the store subdirectory). |
-| `season` | `int` |  | Season END year (2024 = 2023-24), matching the store layout. |
+| `season` | `int` |  | Season **START** year (2023 = 2023-24) -- the key this half of the store is filed under; see the warning above. |
 | `variant` | `Optional[str]` | `None` | Capture variant slug, or `None` for an unparameterized capture. |
 | `result_set` | `Optional[str]` | `None` | Named result set to return when the payload carries several; defaults to the first frame that parses. |
 | `raw_store_dir` | `RawStoreDir` | `None` | Store root spec (dir or URL base) or per-endpoint mapping; `None` falls back to the env vars, `""` disables. |
--- a/sportsdataverse/nba/nba_box_logs.py
+++ b/sportsdataverse/nba/nba_box_logs.py
@@ -11,6 +11,11 @@ from sportsdataverse.nba.nba_stats import nba_stats_leaguegamelog
 # per-100 feature columns (snake-cased leaguegamelog player fields)
 _STATS = ["pts", "fg3m", "fga", "fta", "ast", "oreb", "dreb", "stl", "blk", "tov", "pf"]
 
+#: On-court players per team. A team-game's ``min`` sums all five slots, so
+#: dividing it by this recovers the game's elapsed minutes -- the basis a
+#: player's share of team possessions must be taken against.
+ON_COURT_PLAYERS = 5.0
+
 
 def box_features(
     player_logs: pl.DataFrame,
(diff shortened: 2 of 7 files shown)


saiemgilani:
Added a third fix (`73a5b673`) — found while validating the full 30-season build, and it's the most consequential of the three.

## Legacy pbp reports a literal `"0"` score on non-scoring rows

Pre-2018 play-by-play populates `score_home`/`score_away` only on **scoring** rows and leaves the literal string `"0"` on rebounds, misses and substitutions. Modern games carry the running total on every row.

| | 2015 game | 2019 game |
|---|---|---|
| rows with `scoreHome == "0"` | **320** (Rebound 79, Missed Shot 73, Substitution 49) | 1 |

`"0"` is a **truthy string**, so the forward-fill accepted it and reset the running score to zero on those rows. Possession points are a score *difference*, so differencing against a collapsed score produced possessions worth up to **±140 points** — 15,122 of 37,784 sampled 2014-15 possessions (**~40%**).

That destroyed the RAPM fit built on them:

| | pre-2018 | 2018+ |
|---|---|---|
| RAPM response std | **38.7** (range -140..139) | 1.2 (-3..7) |
| qualified-player RAPM | **-233 .. 341** | -5.9 .. 7.6 |

Every season through 2017 was affected; 2018+ was untouched. Nothing raised.

**Fix:** a basketball score never decreases, so a backward step is ignored — resolving both dialects without detecting which one a game is in. Verified on a real 2014-15 game: points now `[0,3]`, **zero** impossible rows, totalling 201 against the game's actual 101-100 final.

## On the test

My first regression test went through `build_possessions` and **passed against the buggy code** — worse than no test. I caught it by deliberately reverting the fix. The fill is now extracted to `_forward_fill_scores` and tested directly; both tests were verified to fail on the old behaviour and pass on the new.

## `PIPELINE_VERSION` 3 → 4

Every cached possession parquet holds the corrupted points, and the cache keys on `(game_id, PIPELINE_VERSION)` — without the bump a rebuild silently reuses them, here and for every other consumer.

`702 passed`; ruff + mypy clean.

PR #312, opened 2026-08-01
docs(cfb): describe 957 loader return-table columns (2407 deferred -> 1450)

saiemgilani:
Follows #310, which gave generated loader returns tables a `description` column and exposed ~5,850 loader columns to the coverage ratchet — 2,407 of them uncovered. This describes **957 of those (40%)**, all CFB.

| batch | cols | how it was grounded |
|---|---|---|
| `team_summaries` + `_weekly` | **757** | producer arithmetic — `cfb_data_build/team_summaries.py::_summarize_team` |
| `adv_team` + `_gamelog` | **131** | empirical profile of the published 2024 asset |
| `adv_situational` | **71** | empirical profile |

Loader deferred **2,407 → 1,450**. Residual ratchet stays **0**.

## These are grids, not 2,407 independent concepts

`team_summaries` is `{base}_{off|def|margin}[_{pass|rush}][_rank]` — ~28 base metrics generating 378 columns, 179 of them pure `_rank` derivations. So the descriptions are **composed from a verified vocabulary**, and the three generators are committed (`tools/codegen/gen_cfb_*_descriptions.py`) so the derivation stays reproducible and auditable rather than becoming opaque YAML. They enumerate `loader_schemas.yaml`, so re-running is idempotent.

## Grounding came from source, not column names

- **Base metrics** are transcribed from the producer (`EPAdrive = TEPA/n_drives`, `EPAgame = TEPA/n_games`, `success = mean(epa_success)`), not inferred.
- **Rank direction was verified**: offense ranks `ascending=False`, defense `ascending=True`, and lower-is-better metrics flip again — so **rank 1 = best on both sides of the ball**. Documenting that backwards would have been a docs lie in 179 places.
- **`start_position_margin`** uses the special `(100-off) - (100-def)` form, so it is described as field position gained rather than yards-to-goal.
- **Adjusted EPA / strength / valid_games** come from `cfb_adjusted_epa.py` — a ridge (RAPM-style) opponent adjustment fit in-sample, so the text says descriptive-not-predictive.

## The ESPN blocks needed empirical profiling — their names lie

`adv_team` / `adv_situational` are raw ESPN `advBoxScore` fields with no producer arithmetic to read, so each column was profiled against real published data first. That caught three things naming alone would have gotten wrong:

- **The bare `EPA_explosive*` and `EPA_success*` columns are integer play COUNTS** (0–13, 7–56), not EPA totals, despite the `EPA_` prefix. ~40 columns would have been mislabelled.
- **`EPA_overall_off` and `EPA_overall_offense` are exact duplicates** (max abs diff 0). Both now say so, so consumers do not treat them as two signals.
- **`EPA_explosive_rate` is NOT `EPA_explosive / EPA_plays`** — the implied denominator runs 1–27 plays smaller. The text warns that deriving it yourself will not reproduce ESPN value, rather than inventing a formula.

ESPN also emits **both orders** for the same concept (`EPA_success_{situation}` and `EPA_middle_8_success`). That was only caught because the generators **report unparsed columns instead of guessing** — a permissive regex would have left five columns silently blank.

## Nothing was invented

Where a base metric is the mean of an upstream ESPN flag this repo does not define (`havoc`, `explosive`, `opportunity_rate`), the description says "share of plays carrying the flag" instead of asserting a threshold. Anything a generator could not ground is left **blank and reported** — the handful remaining (`season`, `conference`, `game_id`, `week`) resolve through the R-package dict instead.

## Gates

- `generate.py --check` → all generated files current
- `ruff format` + `ruff check` → clean
- `tests/codegen/test_manual_descriptions.py` → **6 passed**, including the filler-quality gate and the orphan-key gate
- residual ratchet **0**, unchanged

## Remaining

1,450 deferred, and the character changes from here — `load_cfb_pbp` (237), `wnba_stats_*` (438), `phf` (114), plus a ~660-column tail across ~80 loaders. Those need genuinely different grounding, so they are better reviewed separately.

## Summary by Sourcery

Document ESPN CFB advanced and team summary loader columns with grounded, empirically verified descriptions, and add generators to keep these descriptions reproducible from source schemas.

Enhancements:
- Introduce codegen scripts to synthesize cfb team_summaries, adv team, and situational column descriptions directly from loader_schemas.yaml, ensuring idempotent, auditable documentation generation.
- Ensure unparsed or ungrounded columns are left blank and reported by the generators rather than described heuristically.

Documentation:
- Fill previously blank descriptions for CFB ESPN advanced team, team gamelog, situational, and team_summaries weekly loader columns, clarifying units, semantics, and relationships (counts vs rates vs EPA).
- Document adjusted EPA, opponent strength, available-yards, and margin/rank metrics in team summaries assets with consistent wording and best-direction notes.
- Clarify quirks in ESPN advBoxScore data such as duplicate overall EPA fields and non-EPA EPA_explosive* counters.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected college football rushing opportunity, highlight-run, and adjusted rushing yardage metrics, including touchdown and fumble-recovery rushes.

* **Documentation**
  * Expanded loader field definitions across college football, hockey, basketball, baseball, and women’s basketball datasets.
  * Added detailed descriptions for play-by-play, advanced statistics, schedules, rosters, ratings, projections, standings, and team summaries.
  * Documented more than 1,300 loader columns and clarified metric calculations and field semantics.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 **Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
 
 - [Unreleased](#unreleased)
+  - [CFB — `opportunity_run` corrected, un-degenerating `opp_highlight_yards` (BREAKING)](#cfb--opportunity_run-corrected-un-degenerating-opp_highlight_yards-breaking)
+  - [CFB / PHF — 1,308 loader return-table columns described](#cfb--phf--1308-loader-return-table-columns-described)
   - [CFB — `team_id` canonicalized to `Int64` at the loader boundary](#cfb--team_id-canonicalized-to-int64-at-the-loader-boundary)
   - [Docs — loader returns tables now carry column descriptions](#docs--loader-returns-tables-now-carry-column-descriptions)
   - [CFB — `adv_*` `pos_team` now holds the team NAME, id moves to `pos_team_id` (BREAKING data change)](#cfb--adv_-pos_team-now-holds-the-team-name-id-moves-to-pos_team_id-breaking-data-change)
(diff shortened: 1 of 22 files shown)


saiemgilani:
All CodeRabbit findings addressed in `784b5b7a`. Five applied, one corrected rather than applied.

**Applied — all real defects in my own work:**

1. **`join_participants=True` in the offline test.** Correct and the most serious of these. The monkeypatch only replaces `cfb_pbp.download`, while the participants path makes its own CDN sidecar and `$ref` fetches, so my "offline" regression test could reach the network. Set to `False`; nothing under test depends on participant joins.

2. **`"on the next 2 plays play"`.** Real grammar bug that shipped into the rendered docs. Fixed at the composer — and separately force-corrected the two already-merged entries, because the additive merge protects existing entries, so a generator fix does not propagate on its own. That is worth knowing for future passes.

3. **`rushing_power_success` labelled a rate.** Verified: `Int64`, range 0–7 on the published 2024 asset — a count. It was falling through to the R-package dict, which rendered "Rushing power success rate". Same class of error as the `EPA_*` columns that turned out to be counts. Now has an explicit entry (manual entries take precedence over the fallback), along with the seven sibling columns you flagged as absent from `DESCS`.

4. **`rushing_highlight_yards` duplicating the per-opportunity wording.** Also correct — that phrasing belongs to `rushing_highlight_yards_per_opp`. This column is the total (1.0–252.0). Fixed.

5. **`unparsed` declared but never populated in the PHF generator.** Good catch: unresolved columns were being silently dropped, which is the exact opposite of the report-never-invent rule the other generators follow. It now reports them.

6. **Stale generator docstring.** Described the `opportunity_run` inversion as present-tense fact after the fix landed. Now past tense.

**Corrected rather than applied — `rink_id` / `rink` / `external_url` / `highlight_color`:**

These are not a mistyped source schema. Checked against the published asset: all four are **all-null in every season** (0 non-null of 17 rows in 2021), and `Boolean` is simply what polars infers for an entirely empty column. Changing the declared type would make the docs disagree with the actual asset. The columns now state that they are never populated and why the dtype looks the way it does — which is the information a consumer actually needs.

**Gates:** `generate.py --check` current, `ruff` clean, `test_manual_descriptions` + `test_cfb_opportunity_run` → 10 passed.

PR #315, opened 2026-08-01
fix(codegen): re-derive 6 stale loader schemas; finish the loader column descriptions

saiemgilani:
Finishes the work from #312. **Loader return-table columns: 0 undocumented, down from 2,407.**

## Half the remaining work was fiction

65 of the 127 supposedly-undocumented columns **do not exist in the published data** — declared in `loader_schemas.yaml` but never shipped. Describing them would have documented columns no consumer can receive.

The drift is **bidirectional**, so these tables were not merely incomplete — they were wrong:

| loader | declared | ships | phantom | missing from declared |
|---|---|---|---|---|
| `load_nhl_player_boxscore` | 50 | 37 | **46** | **33** (only ~4 overlap) |
| `load_nhl_schedule` | 28 | 35 | 16 | 23 |
| `load_nhl_pbp` | 103 | 94 | 35 | 26 |
| `load_nhl_team_boxscore` | 18 | 19 | 8 | 9 |
| `load_wnba_stats_coaches` | 12 | 11 | 2 | 1 |
| `load_wnba_player_boxscore` | — | 57 | 3 | — |

All six are re-derived from a diagonal union of real published seasons, fixing both directions at once.

**Why the audit had to probe per-season:** `load_wnba_player_boxscore.fg/fg3/ft` were reported *present* by a multi-season union, but probing 11 seasons individually found them in **none**. The aggregate lied; direct probing did not.

## The last 62 descriptions

NHL localized name keys use the framing verified against the published 2024 penalties asset, not the intuitive one:

- `lastName.cs` — `Hagg` → `Hägg`: **adds** diacritics the default folds to ASCII
- `lastName.sv` — both directions: `Holmstrom`→`Holmström` *and* `Hakanpää`→`Hakanpaa`
- `firstName.es` — `Aliaksei`→`Alexei`, `Joshua`→`Josh`: a **different name**, not a re-spelling

So these are not "locale spellings", which is what an unverified first pass claimed.

Four columns are flagged as carrying **no signal** rather than described as if they did: `linescore` and `espn_sourced` are constant true, `draft_type` is always `"Draft"`, `season_2` duplicates `season` on every row.

## One self-inflicted bug worth recording

The re-derivation emitted `type: Null` for an all-null column, and **YAML parses a bare `Null` as the null literal** — so `c["type"]` came back as `None` and crashed codegen in `_build_loader_docstring`. Quoted to `'Null'`. A dtype whose name collides with a YAML keyword is a genuine trap.

## Scope note

[...]

PR #316, opened 2026-08-01
feat(cfb): loaders for the two published FPI datasets

saiemgilani:
Companion to [cfbfastR-cfb-data#29](https://github.com/sportsdataverse/cfbfastR-cfb-data/pull/29), which published both datasets.

| loader | shape (2024) | grain |
|---|---|---|
| `load_cfb_fpi_weekly` | `2278 x 51` | team × season_type × **week** |
| `load_cfb_power_index` | `1833 x 7` | game × team |

## `power_index` was a link list

It previously declared `[$ref, game_id, season, week]` because the published asset only carried unresolved `$ref` links. It now ships the real matchup predictions — `teampredptdiff`, `gameprojection`, `matchupquality`, `teamadjgamescore` — so the declared schema is re-derived from the published data.

The stale `$ref` description is dropped; the orphan-key test caught it, which is exactly what that gate is for.

## Two leakage flags come through on `fpi_weekly`

`snapshot_out_of_sequence` and `snapshot_is_contemporaneous` are carried on the asset and surface through the loader. They matter: ESPN's **week 1 slot is a post-season overwrite** (2019 wk1 is stamped *April 2020*), and **seasons before 2015 are retrospective backfills** (2012 = 0% contemporaneous, 2015 = 94%, 2024 = 100%). Without them a modeller would silently use end-of-season information as an as-of-week-1 rating.

## Both declare `id_int64: [team_id]`

Matching the canonicalization the rest of the CFB surface uses, so joins against the box/pbp/adv families work without a cast.

## Guard for a trap that has now bitten twice

polars' all-null dtype is literally `Null`, and **YAML parses a bare `Null` as the null literal** — so a schema re-derived from an all-null column writes `type: Null`, reads back as `None`, and blows up much later in `_build_loader_docstring` with an opaque `len(None)`. That happened on the NHL re-derivation in #315 and again here on `fpi_weekly`.

`_assert_declared_types_are_strings` now converts the delayed crash into an actionable message naming the loader and column. Verified it fires on a null dtype and passes clean schemas.

Placing it also exposed a second bug worth recording: inserting the function immediately above `_loader_schemas` **silently stole that function's `@lru_cache` decorator**, which then raised `unhashable type: dict` on the first call. The decorator is back where it belongs.

## Gates

- `generate.py --check` current, `ruff` clean
- `tests/codegen/test_manual_descriptions.py` + `test_id_casts.py` → **18 passed**
- both loaders verified against the live release

## Summary by Sourcery

Add a new weekly ESPN FPI loader for CFB releases, refresh the power index loader and schema to expose actual prediction fields and team IDs, and harden codegen against null dtype schema regressions.

New Features:
- Introduce the load_cfb_fpi_weekly loader and wire it into the CFB API, release endpoints, and documentation.

Bug Fixes:
- Correct the espn_cfb_power_index schema and loader to remove stale $ref/week fields and align with the published matchup prediction data, including team_id casting.
- Prevent silent YAML/Polars null dtype issues from causing late failures in loader docstring generation by validating declared types during codegen.

Enhancements:
- Declare team_id as an Int64 identifier for both FPI weekly and power index loaders to ensure consistent cross-dataset joins.
- Expand CFB loader reference docs to cover the new FPI weekly dataset and updated power index columns.
- Update the documented count of CFB dataset loaders to reflect the additional loader.

## Summary by CodeRabbit

- **New Features**
  - Added access to weekly college football FPI data for seasons from 2005 onward.
  - Supports optional pandas output and skips unavailable seasons with warnings.
  - Expanded power index data with detailed game, team, and projection fields.

- **Documentation**
  - Added loader usage examples, schemas, and automation-status details.
  - Updated CFB documentation to include the new dataset.

- **Bug Fixes**
  - Standardized team identifiers as nullable 64-bit integers.
  - Added validation for incomplete schema type definitions.


saiemgilani:
Fixed in `596c66e3` — correct catch, and a good one.

The guard was added to `_loader_schema_table` while its own docstring names `_build_loader_docstring` as the crash site. That function reads `_loader_schemas()` directly and indexes `c["type"]` with no validation, so **the guard did not protect the thing it existed to protect** — a future all-null column would still have produced the opaque `len(None)` it was written to prevent.

Moved into `_loader_schemas()` itself, the single chokepoint every consumer goes through. The per-call-site invocation is now redundant and removed.

**Verified rather than assumed:** reintroduced a bare `type: Null` into the yaml and confirmed codegen now fails immediately with

```
loader_schemas.yaml has null dtypes (quote them as 'Null'): load_cfb_player_box.jersey
```

instead of crashing deep in docstring rendering. Restored afterwards.

Gates: drift clean, ruff clean, 18 codegen tests pass.

PR #317, opened 2026-08-02
docs(cfb): describe the FPI loader columns from ESPN's own field text

saiemgilani:
The two FPI loaders added in #316 brought **48 columns with no returns-table description**, taking the coverage ratchet back off zero. This closes it.

## Most of the text is ESPN's, not mine

ESPN ships a `description` beside every FPI field in its own payloads. The generator fetches both — the weekly table's `predictives`/`efficiencies` blocks and the per-event `stats` block — and attaches ESPN's wording to the matching column. That covers **43 of 48**. Writing my own prose for `gamecontrol` or `accomplishment` would only drift from what the provider means by them.

## The rest are authored, each grounded in the published data

| column | what the data shows |
|---|---|
| `projectedt` | **null in all 23,641 rows** — ESPN emits the key beside `projectedw`/`projectedl` but never a value. Ties were abolished in 1996. Hence the `Null` dtype. |
| `adjwinpctrank`, `projectedwpctrank` | ranks whose underlying percentage ESPN **does not publish** (there is no `adjwinpct` column). `0` is an unranked placeholder, not a rank — it appears where the underlying value is null. |
| `run_date_time_key` | the **as-of** date, which is not `last_updated` (**when computed**). The gap between them is the retrospection. |
| `snapshot_out_of_sequence` / `_is_contemporaneous` | the two point-in-time traps, documented where a consumer meets them rather than only in a module docstring. |

## Four of ESPN's own descriptions are filler

`"Won Games."`, `"Lost Games."`, `"Tied Games."`, `"Rank."` — the `test_no_filler_descriptions` gate rejects a description that only restates the column name, and it was right to. Replaced with text that answers the actual question here, which is *which* of several similar columns this one is: `numwins` is the **actual** record, not `projectedw` (projection) and not `adjwins` (opponent-adjusted).

## The rank caveat is measured, not asserted

I nearly wrote "`fpirank` looks stale" off one 2015 row. Instead I derived the rank implied by the published `fpi` within each snapshot and compared:

```
disagreeing rows: 140 of 23,591 (0.6%)
  rank    matches derived: 140
  fpirank matches derived:   0
agreeing rows that match derived: 23,451 of 23,451
```

So `rank` is authoritative and `fpirank` is stale on exactly those rows. Both columns now say to prefer `rank`.

## Drive-by fix in the merge helper

`cur.get(k, {})` returned `None` for a loader present in the yaml with an **empty block** — the two-arg default only covers a *missing key*, not a *null value*, and YAML parses an empty block to `None`. It crashed the merge.

Worth noting how it hid: the crash was invisible through `cmd 2>&1 | tail`, because a pipeline's exit code is `tail`'s, not the failing command's — so the `&&` chain marched on and the next step 'succeeded' against unmerged input. Caught by checking the ratchet count, not the exit code.

## Verification

- ratchet: `residual = 0`, `loader deferred = 0` (was 48)
- rendered: **51** and **7** rows, **0 blank** — confirmed by parsing the built page, after a first probe that matched the wrong heading format and cheerfully reported "0 blank" because it had found nothing at all
- generator is idempotent: re-run merges `+0`
- `tests/codegen/test_manual_descriptions.py` 6 passed; ruff clean; `generate.py --check` current

## Summary by Sourcery

Document ESPN FPI loader columns using provider-sourced descriptions and a small set of authored explanations, and fix the column-description merge helper to handle empty YAML blocks safely.

Bug Fixes:
- Correct the merge_column_descriptions helper to treat null YAML loader blocks as empty mappings, preventing crashes when merging new descriptions.

Enhancements:
- Introduce a dedicated generator script that pulls FPI field descriptions from ESPN APIs, augments them with authored caveats, and emits a reusable YAML for manual column descriptions.

Documentation:
- Add full, user-facing descriptions for all columns in the cfb_fpi_weekly and espn_cfb_power_index loaders, clarifying semantics, rank behavior, and point-in-time caveats.

## Summary by CodeRabbit

* **Documentation**
  * Expanded college football FPI loader documentation with descriptions for metrics, rankings, projections, probabilities, schedule strength, efficiency, and snapshot metadata.
  * Added definitions for power index game projections, matchup quality, adjusted game scores, and predicted point differentials.
  * Clarified field meanings, snapshot timing, and result ordering for easier interpretation.

* **Chores**
  * Improved automated generation and merging of column descriptions, including support for empty metadata entries and more consistent coverage.

### DIFF
--- a/docs/docs/cfb/reference/loaders.md
+++ b/docs/docs/cfb/reference/loaders.md
@@ -965,52 +965,52 @@ Release: [cfb_fpi_weekly](https://github.com/sportsdataverse/sportsdataverse-dat
 | `week` | Int64 | Game week of the season. |
 | `team_id` | Int64 | ESPN team id. |
 | `last_updated` | String | Timestamp ESPN last refreshed the power index. |
-| `run_date_time_key` | Int64 |  |
-| `snapshot_out_of_sequence` | Boolean |  |
-| `fpi` | Float64 | ESPN Football Power Index rating (projected scoring margin vs. average team). |
-| `fpirank` | Float64 |  |
-| `projectedw` | Float64 |  |
-| `projectedl` | Float64 |  |
-| `projectedt` | Null |  |
-| `projectedwpctrank` | Float64 |  |
-| `probwinout` | Float64 |  |
-| `probwinconf` | Float64 |  |
-| `sosremainingrank` | Float64 |  |
-| `accomplishment` | Float64 |  |
-| `accomplishmentrank` | Float64 |  |
-| `adjwins` | Float64 |  |
-| `adjlosses` | Float64 |  |
-| `adjwinpctrank` | Float64 |  |
-| `gamecontrol` | Float64 |  |
-| `gamecontrolrank` | Float64 |  |
-| `adjavgingamewp` | Float64 |  |
-| `adjavgingamewprank` | Float64 |  |
-| `avgingamewp` | Float64 |  |
-| `avgingamewprank` | Float64 |  |
-| `avgsosrank` | Float64 |  |
-| `topsosrank` | Float64 |  |
-| `epaoffense` | Float64 |  |
-| `epadefense` | Float64 |  |
-| `epaspecialteams` | Float64 |  |
-| `probwindiv` | Float64 |  |
-| `probmakeplayoffs` | Float64 |  |
-| `probmaketitlegame` | Float64 |  |
-| `numwins` | Float64 |  |
-| `numlosses` | Float64 |  |
-| `numties` | Float64 |  |
-| `probwintitle` | Float64 |  |
-| `rankchange7days` | Float64 |  |
-| `prob6wins` | Float64 |  |
-| `rank` | Float64 | Position of the school within the poll for the given week (1 = top-ranked). |
-| `offefficiency` | Float64 |  |
-| `offefficiencyrank` | Float64 |  |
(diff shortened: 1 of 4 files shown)


saiemgilani:
Partly taken in `3c289fa5`.

**Took the useful half.** Both functions now document what is actually non-obvious: that `espn_descriptions` keys on ESPN's raw field names — which *are* the published column names, so the result is lookup-by-column — that it merges two payloads because neither alone covers every column, and that `main` returns 0 even with uncovered columns because the ratchet, not this script, decides what counts as undescribed.

**Skipped `Args:` / `Raises:` / `Example:`.** The Google-style convention in CLAUDE.md scopes to the package's public callables, where docstrings feed runtime `help()` and the generated reference pages. This is a build script — no reference page, no external caller, invoked once by a maintainer. Checked rather than assumed:

- no ruff `D` rules configured (`select` is unset)
- no test asserts docstrings over `tools/`
- every sibling generator (`gen_cfb_adv_descriptions`, `gen_cfb_rest_descriptions`, `gen_cfb_pbp_descriptions`) has **no docstring at all** on `main`/`describe` — this file was already the most documented in the directory

`Args: None.` is also precisely the filler this repo's own `test_no_filler_descriptions` gate rejects for column text, and an `Example:` on `main()` would be a runnable block that fires two live ESPN requests as a side effect of reading the docs.

Ruff clean, generator re-runs identically.

PR #365, opened 2026-08-13
refactor(nba): repoint the _v3 loaders onto the production release tags

saiemgilani:
## What

Repoints the three NBA `_v3` loaders off the release tags being retired
(`nba_stats_pbpv3`, `nba_stats_possessions_v3`, `nba_stats_lineups_v3`) and onto
the production tags. **No tag or asset is retired or deleted here** — this is
repointing only, so the tags can be retired separately.

## Superset verification (done before changing anything)

Each `_v3` tag holds one season-2025 asset, built before the 2025-26 Finals
ended. I read both live assets and compared game-id sets directly:

| dataset | `_v3` rows / games | production rows / games | v3 games missing from prod |
|---|---|---|---|
| pbp | 704,314 / 1,394 | **707,440 / 1,400** | **0** |
| possessions | 277,222 / 1,394 | **278,341 / 1,400** | **0** |
| lineups | 704,314 / 1,394 | **707,440 / 1,400** | **0** |

The six extra games are exactly `0042500317`, `0042500401`–`0405`. Prefix
breakdown confirms every game type is admitted:

```
v3   {'001': 71, '002': 1230, '003': 7, '004': 79, '005': 6, '006': 1}
prod {'001': 71, '002': 1230, '003': 7, '004': 85, '005': 6, '006': 1}
```

Two checks beyond game-id containment, because "same games" is not "same data":

- **Row-level:** on the 1,394 shared games, **0** games have fewer rows in
  production than in `_v3`.
- **Column-level:** `_v3` has **no** column production lacks; production adds
  exactly one (`season`); zero dtype divergence.

## Which option I chose: (b) deprecate

`load_nba_stats_*_v3` become `DeprecationWarning` shims (removal `0.1.0`)
forwarding to the production loaders. They keep working.

Rationale — (b) over (a):

1. The repo has a **formal deprecation policy** (`sportsdataverse/_deprecation.py`)
   and an established precedent (`load_nfl_ngs_*`, `load_nfl_pfr_advstats_*`).
   CLAUDE.md names it as the pattern. (a) would leave three permanently
   misleading names promising a "v3" dataset that no longer exists distinctly.
2. Post-repoint these are **exact functional duplicates** of the production
   loaders — same season convention, same columns, same data. Keeping two names
   for one dataset forever is the cost (a) never stops paying.
3. Nothing is silently deleted; callers get a warning naming the replacement.

| Deprecated | Forwards to | Release now read |
|---|---|---|
| `load_nba_stats_pbp_v3` | `load_nba_stats_pbp` | `nba_stats_pbp` |
| `load_nba_stats_possessions_v3` | `load_nba_stats_possessions` | `nba_stats_possessions` |
| `load_nba_stats_lineups_v3` | `load_nba_stats_game_lineups` | `nba_stats_game_lineups` |

`lineups_v3` maps to **`game_lineups`**, not `lineups` — `nba_stats_lineups` is
the leaguedash lineup cube, a different dataset (different shape, `min_season`
2007). Asset sizes corroborate: `nba_lineups_v3_2025` 1,039,847 B vs
`nba_lineups_2026` 1,051,800 B.

## Season convention — unchanged (the highest-risk part)

**Every one of these loaders exposed a START-year `seasons` argument before this
change and still does.** The `_v3` assets were START-year keyed
(`play_by_play_v3_2025` = 2025-26) and production is END-year keyed
(`nba_play_by_play_2026` = 2025-26), but that translation lives in the codegen
asset-path template (`{season + 1}`), **not** at the shim boundary. So the shims
forward `seasons` **untouched, with no arithmetic**:

| loader | `seasons` before | `seasons` after | 2025 resolves to |
|---|---|---|---|
| `load_nba_stats_pbp_v3` | START year | **START year** | `nba_play_by_play_2026.parquet` (2025-26) |
| `load_nba_stats_possessions_v3` | START year | **START year** | `nba_possessions_2026.parquet` (2025-26) |
| `load_nba_stats_lineups_v3` | START year | **START year** | `nba_lineups_2026.parquet` (2025-26) |

Tests pin each shim to the same asset as its target for 1996 / 2010 / 2025, so a
future off-by-one fails rather than silently shifting every caller by a year.

## Live verification

Actually called, asserting **populated** output (these loaders `cli_warn`-and-skip
a missing season, so "no exception" proves nothing):

```
load_nba_stats_pbp_v3(seasons=2025)          rows=707,440  games=1400  cols=49
  DeprecationWarning: ... use load_nba_stats_pbp instead.
  prefixes {'001': 71, '002': 1230, '003': 7, '004': 85, '005': 6, '006': 1}
  vs retired _v3: +3,126 rows, +6 games   shim output == target  OK
load_nba_stats_possessions_v3(seasons=2025)  rows=278,341  games=1400  cols=35
  vs retired _v3: +1,119 rows, +6 games   shim output == target  OK
load_nba_stats_lineups_v3(seasons=2025)      rows=707,440  games=1400  cols=14
  vs retired _v3: +3,126 rows, +6 games   shim output == target  OK
```

## Also here: the `season` COLUMN divergence (undocumented until now)

A parallel sdv-db investigation surfaced that the four `{season + 1}`-shifted NBA
loaders return frames stamped with the **asset's** year, so the `season` column
does **not** equal the `seasons` argument. I confirmed this independently against
live assets rather than taking it on report:

```
pbp / schedules / possessions / game_lineups (shifted)   season col = [2026]
team_boxscores / officials      (unshifted)              season col = [2025]
```

Both are the 2025-26 season. This was silent, and it is load-bearing: a
partitioner keying off the column writes 2024-25 **over** the 2025-26 partition.

**Decision: document, do not restamp.** The column is the published asset's own
identity; rewriting it would make the frame disagree with the file it was read
from and destroy provenance for exactly the debugging this causes. sdv-db has
already shipped `key_on_requested_season` (#36) on the assumption the column stays
END-year — restamping would re-break what was just fixed. Restamping is also a
silent mutation of a column callers may already join on.

Applied at the single token-driven point in the docstring builder, so it covers
all four shifted loaders **and** the three shims automatically and cannot drift
from the asset path it describes. A new codegen test fails if a future shifted
loader ships without the note (mutation-tested — it catches all 7 when the text
is removed).

## Implementation note

Added a `deprecated_for:` field to the loader codegen (`spec.py` + template +
`generate.py`) rather than hand-editing a generated module — `nba_loaders.py` is
`DO NOT EDIT`. The other seven `*_loaders.py` diffs are one added import line
each from the shared template.

## Verification

- `generate.py --check` — green
- `tests/codegen/` — 157 passed, 11 skipped
- `tests/codegen/test_loader_season_convention.py` — 29 passed (was 16)
- `ruff check` / `ruff format --check` — clean
- `nba_loaders.py` is not in the mypy ratchet

Not merging.

## Summary by Sourcery

Deprecate the three NBA `*_v3` stats loaders in favor of their production counterparts and document the season-year semantics for NBA loaders whose assets use an end-year key, backed by codegen and tests to enforce the behavior and deprecation policy.

Enhancements:
- Introduce codegen support for deprecated loaders via a `deprecated_for` field, generating shims that forward to replacement loaders while emitting standardized deprecation warnings.
- Update NBA `_v3` loaders (`pbp_v3`, `possessions_v3`, `lineups_v3`) to act as pass-through shims to the corresponding production loaders and extend their supported season range via shared asset paths.

Documentation:
- Document in NBA loader docstrings and changelog that the `season` column reflects the asset END year, not the requested `seasons` START year, and warn users against joining across shifted and unshifted loaders without normalization.
- Refresh NBA loader reference docs and CHANGELOG entries to show that the `_v3` loaders now read production releases and include the `season` column in their schemas.

Tests:
- Add tests ensuring `_v3` NBA loaders forward to the same assets as their production counterparts, emit `DeprecationWarning` naming the replacement, and that no loader still references retired `*_v3` release tags.
- Add a codegen test that requires all season-shifted loaders with a `season` column to include the documented column divergence warning in their docstrings.

Chores:
- Import the shared deprecation helper into all league loader modules so future generated shims can rely on a single `warn_deprecated` implementation.

## Summary by CodeRabbit

- **Deprecations**
  - NBA lineups, play-by-play, and possessions V3 loaders now warn and forward requests to their recommended replacements.
  - Deprecated loaders support seasons from 1996 onward while preserving start-year season inputs.

- **Data Updates**
  - Returned `season` values for shifted NBA datasets now clearly reflect the asset’s end year.

- **Documentation**
  - Updated NBA loader references, schemas, release links, migration guidance, and deprecation notices.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 **Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
 
 - [Unreleased](#unreleased)
+  - [Deprecated — the three NBA `*_v3` loaders now read the production releases](#deprecated--the-three-nba-_v3-loaders-now-read-the-production-releases)
+  - [Docs — NBA shifted loaders: the `season` COLUMN is the END year](#docs--nba-shifted-loaders-the-season-column-is-the-end-year)
   - [New — `sportsdataverse.wexp`: win-expectancy bake-off harness (NFL + CFB)](#new--sportsdataversewexp-win-expectancy-bake-off-harness-nfl--cfb)
   - [New — `load_nfl_ratings_weekly`: per-week as-of NFL ratings vintages](#new--load_nfl_ratings_weekly-per-week-as-of-nfl-ratings-vintages)
   - [New — `sportsdataverse.scrape.espn`: shared ESPN `-raw` archive engine](#new--sportsdataversescrapeespn-shared-espn--raw-archive-engine)
@@ -230,6 +232,54 @@
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
 ## Unreleased
+
+### Deprecated — the three NBA `*_v3` loaders now read the production releases
+
+`load_nba_stats_pbp_v3`, `load_nba_stats_possessions_v3` and
+`load_nba_stats_lineups_v3` read one-asset, season-2025-only release tags
+(`nba_stats_pbpv3`, `nba_stats_possessions_v3`, `nba_stats_lineups_v3`) that are
+being retired. Those assets were built before the 2025-26 Finals ended and are
+missing six games (`0042500317`, `0042500401`–`0405`).
+
+The Program V pipeline now admits every game type (preseason `001`, regular
+`002`, All-Star `003`, playoffs `004`, play-in `005`, NBA Cup `006`) across all
+30 seasons, so the production tags carry a **strict superset** — verified live
+for the overlapping season: 707,440 rows / 1,400 games vs 704,314 / 1,394, with
+zero games lost, zero games short a row, and no column dropped (production adds
+`season`).
+
+All three keep working and are now **`DeprecationWarning` shims** (removal in
+`0.1.0`) forwarding to their production successors:
+
+| Deprecated | Use instead | Release now read |
+|---|---|---|
+| `load_nba_stats_pbp_v3` | `load_nba_stats_pbp` | `nba_stats_pbp` |
+| `load_nba_stats_possessions_v3` | `load_nba_stats_possessions` | `nba_stats_possessions` |
+| `load_nba_stats_lineups_v3` | `load_nba_stats_game_lineups` | `nba_stats_game_lineups` |
+
+**The `seasons` argument is unchanged.** It was the season's START year before
+and still is: `load_nba_stats_pbp_v3(seasons=2025)` meant 2025-26 when it read
+`play_by_play_v3_2025.parquet` and still means 2025-26 now that it reads
+`nba_play_by_play_2026.parquet`. The retired assets were START-year keyed and the
+production assets are END-year keyed; that translation lives in the loader's
+`{season + 1}` asset-path template, so the shims forward `seasons` untouched.
+
+### Docs — NBA shifted loaders: the `season` COLUMN is the END year
+
+The four NBA loaders whose asset path carries `{season + 1}`
+(`load_nba_stats_schedules`, `load_nba_stats_pbp`, `load_nba_stats_possessions`,
+`load_nba_stats_game_lineups`) return frames stamped with the **asset's** year,
+so the `season` column does not equal the `seasons` argument:
+`load_nba_stats_schedules(seasons=2024)` returns rows reading `season == 2025`.
+Unshifted NBA siblings (`team_boxscores`, `officials`, `rosters`) stamp `2024`
+for that same real season.
+
+This was undocumented and is load-bearing — a partitioner keying off the column
+writes the 2024-25 season over the 2025-26 partition. The divergence is now
+documented in all four docstrings (and the three shims that inherit it), with a
+codegen test that fails if a future shifted loader ships silent. The column is
+**documented, not restamped**: it is the published asset's own identity, and
+rewriting it would make the frame disagree with the file it came from.
 
 ### New — `sportsdataverse.wexp`: win-expectancy bake-off harness (NFL + CFB)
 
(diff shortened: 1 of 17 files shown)


saiemgilani:
Thanks both — addressed in 03f9fb9.

**CodeRabbit — conditional `warn_deprecated` import: taken.** Good catch, and it
made the diff strictly smaller. The import is now emitted only for modules that
actually render a `deprecated_for` shim, which drops both the unused import and
the blanket `noqa`. The seven unrelated loader modules (`cfb`, `mbb`, `mlb`,
`nhl`, `pwhl`, `wbb`, `wnba`) now match `main` byte-for-byte instead of each
carrying a spurious one-line diff.

**Sourcery — NBA-specific wording in league-agnostic codegen: taken.** The
concrete sibling table names (`team_boxscores`, `officials`, `rosters`) are now
emitted only when `ld.league == "nba"`; any future `{season + N}` loader in
another league gets the same hazard stated generically. It was accurate for
every loader it currently renders on — `test_only_nba_stats_families_carry_the_end_year_offset`
asserts the shifted set is NBA-only — but asserting NBA table names from a
league-agnostic builder is a latent trap, so it's worth removing now.

**Sourcery — `LOADER_DEPRECATION_REMOVED_IN` single source of truth: declining,
with reasoning.** Specifically the suggestion to tie it to the package version.
`removed_in` is a *policy* commitment (which future release removes the API),
not a function of the current version — `sportsdataverse/_deprecation.py`
requires a deprecation to survive at least two minor releases and to name the
removal version up front. Deriving it from `__version__` would silently change
the promise made to users on every version bump, which is the opposite of what
the policy guarantees.

On centralization: a module-level constant in `generate.py` is already *more*
centralized than the existing precedent — the hand-written NFL shims
(`load_nfl_ngs_*`, `load_nfl_pfr_advstats_*`) hardcode `removed_in="0.1.0"` at
each call site. Hoisting a shared constant into `_deprecation.py` so both the
generated and hand-written shims read one value is a reasonable follow-up, but
it means touching ~11 unrelated NFL loaders and is out of scope for a release-tag
repoint.

**Note on the CodeRabbit status check:** on the companion PR
(sportsdataverse/sdv-db#37) the CodeRabbit check reported **SUCCESS** while its
comment said *"Review limit reached — we couldn't start this review."* The
conclusion is not a reliable signal of whether a review actually ran; the comment
body is.

PR #367, opened 2026-08-13
feat(nba_stats)!: point the remaining 12 loaders at END-year-named assets

saiemgilani:
**BREAKING** for anyone fetching the release assets directly by URL. The public `seasons` argument is **unchanged** — still the season's START year for every nba_stats loader.

## Why

The nba_stats release tags were internally inconsistent: **4 END-year named** (`nba_play_by_play_{season + 1}`) against **47 START-year named** (`shots_{season}`, plus the whole 36-table leaguedash cube). One schema, two conventions.

The 2026-08-13 republish moved all **1,931 assets across 14 tags** onto END-year names and rewrote their `season` column to match. These 12 loaders follow it:

`coaches`, `game_rosters`, `lineups`, `officials`, `player_boxscores`, `player_game_logs`, `player_season_stats`, `rosters`, `shots`, `standings`, `team_boxscores`, `team_season_stats`

Verified live against the republished assets:

```
load_nba_stats_coaches(seasons=2025) -> rows=267     season=[2026]
load_nba_stats_shots(seasons=2025)   -> rows=233632  season=[2026]
```

2025-26 is what the `nba` (ESPN) schema has always called 2026, so the two now agree.

## wnba_stats is deliberately untouched

WNBA is single-calendar-year — start and end are the same year, so shifting it would walk it off by one. Worth calling out: **my first cut shifted 8 WNBA URLs by accident**, because `wnba_stats_` contains the substring `nba_stats_`. Reverted and redone with an anchored pattern; the sdv-db side has a test asserting WNBA URLs stay unshifted.

## Docstring correction

The generated `.. warning::` named `team_boxscores`, `officials` and `rosters` as "unshifted siblings" to avoid joining against — all three are shifted by this change, so that guidance had become actively wrong. It now states the whole nba_stats family agrees and points at football (`cfb`/`nfl`) as the remaining START-year convention.

## Related

Pairs with sdv-db's END-year re-key (sdv-db #41) — the database moved to END-year season keys in the same operation, so the loader, the asset and the warehouse now use one convention for NBA.

## Summary by Sourcery

Align nba_stats loaders and documentation with END-year keyed release assets while keeping the public seasons argument as the START year.

Enhancements:
- Update 12 nba_stats loaders to fetch END-year-named parquet assets while translating from START-year seasons internally.
- Unify NBA loader docstrings to state that all nba_stats tables now carry END-year season stamps and to highlight cross-sport START vs END year conventions.

Build:
- Adjust codegen release endpoint templates so generated NBA loaders use END-year-based asset URLs for the nba_stats family.

Documentation:
- Refresh NBA loader reference docs to reflect END-year-based asset filenames for the affected nba_stats releases.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected NBA data loading so starting-year seasons consistently retrieve assets labeled by the following end year.
  * Updated season labels across coaches, rosters, game data, player and team statistics, standings, shots, lineups, officials, and logs.
  * Preserved compatibility for deprecated loader aliases.
  * Confirmed WNBA loaders retain their existing season convention.

* **Documentation**
  * Clarified NBA season conventions, republished dataset coverage, and differences from college football and NFL labeling.

### DIFF
--- a/docs/docs/nba/reference/loaders.md
+++ b/docs/docs/nba/reference/loaders.md
@@ -659,7 +659,7 @@ load_nba_stats_schedules(seasons=2025)
 
 ## `load_nba_stats_coaches`
 
-Release: [nba_stats_coaches](https://github.com/sportsdataverse/sportsdataverse-data/releases/tag/nba_stats_coaches) · asset `https://github.com/sportsdataverse/sportsdataverse-data/releases/download/nba_stats_coaches/coaches_{season}.parquet`
+Release: [nba_stats_coaches](https://github.com/sportsdataverse/sportsdataverse-data/releases/tag/nba_stats_coaches) · asset `https://github.com/sportsdataverse/sportsdataverse-data/releases/download/nba_stats_coaches/coaches_{season + 1}.parquet`
 ### Returns
 
 | col_name | type | description |
@@ -682,7 +682,7 @@ load_nba_stats_coaches(seasons=2025)
 
 ## `load_nba_stats_game_rosters`
 
-Release: [nba_stats_game_rosters](https://github.com/sportsdataverse/sportsdataverse-data/releases/tag/nba_stats_game_rosters) · asset `https://github.com/sportsdataverse/sportsdataverse-data/releases/download/nba_stats_game_rosters/game_rosters_{season}.parquet`
+Release: [nba_stats_game_rosters](https://github.com/sportsdataverse/sportsdataverse-data/releases/tag/nba_stats_game_rosters) · asset `https://github.com/sportsdataverse/sportsdataverse-data/releases/download/nba_stats_game_rosters/game_rosters_{season + 1}.parquet`
 ### Returns
 
 | col_name | type | description |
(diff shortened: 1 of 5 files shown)


saiemgilani:
Addressed in `8e417901`.

**Taken — the `min_season` example (CodeRabbit + Sourcery, same finding).** Real bug, verified: `load_nba_stats_lineups` has `min_season=2007` and raises `SeasonNotFoundError` below it, so the hard-coded `1996` example told users to make a call that errors. `_build_loader_docstring` now derives the example from `ld.min_season` and builds the season label with it, so `lineups` reads *"e.g. `2007` for the 2007-08 season"* while the `1996`-floor loaders are unchanged. Regenerated.

**Taken — the exact `+1` assertion.** Correct: `SEASON_TOKEN` captures the `N` in `{season + N}`, so a truthiness test would have accepted `{season + 2}` as satisfying an END-year contract that means exactly one year. Now compares `== "1"`.

**Already done — adding the 12 loaders to the expected set.** That landed in `7b2e94de`; the review was against the first commit.

**Not taken — factoring the repeated warning block into a shared helper (Sourcery).** It already comes from exactly one place, `_build_loader_docstring`. The repetition is in the *generated* output, which is what generating it is for; extracting a helper from the emitter would add a layer without removing a source of truth.

PR #401, opened 2026-08-27
fix(torvik): reject an HTML body instead of parsing it as a CSV

saiemgilani:
Found while fixing the nightly `wehoop-wbb-data` build, which failed on 2026-08-24 in `team_crosswalk`, `schedule_crosswalk` **and** `player_crosswalk` at once with:

```
polars.exceptions.ColumnNotFoundError: unable to find column "team";
valid columns: ["doctype_html_public_w3c_dtd_html_4_01_transitional_en_http_www_w3_org_tr_html4_loose_dtd",
                "unnamed"]
  ...
  File "sportsdataverse/wbb/wbb_crosswalk.py", line 231, in _assemble_team_crosswalk
    bart2 = bart.select(
```

That column name is a snake-cased HTML DOCTYPE. barttorvik.com answers a transient outage with an **HTML page and HTTP 200**, `torvik_runtime._get` correctly returns the body as text for a non-JSON content type, and `parse_torvik_csv` then read the HTML as a headed CSV — so the DOCTYPE became the only column. The real failure surfaced several frames away, in a function that had nothing to do with it.

The existing guard in `_assemble_team_crosswalk` is `bart is None or bart.height == 0`, which an HTML frame passes: it has rows, just not the right ones.

**Change:** `parse_torvik_csv` raises when the payload is an HTML document, naming what actually arrived. The crosswalk builders already wrap the fetch in `require_source()`, whose docstring states the contract exactly — *"a provider that cannot be produced — the module is missing, the import fails, the request errors, the payload will not render — is a build failure"* — so this routes into machinery that was already designed for it, and the caller gets `CrosswalkSourceError: bart_wbb_ratings(year=2026) ...` instead of a `ColumnNotFoundError`.

The documented zero-row contract is unchanged for genuinely empty input:

```
  PASS  HTML rejected: expected a barttorvik.com CSV data file but received an HTML document...
  PASS  real CSV still parses: (2, 3) cols=[rank, team, conf]
  PASS  empty        -> zero-row frame (contract unchanged)
  PASS  header only  -> zero-row frame (contract unchanged)
  PASS  non-text     -> zero-row frame (contract unchanged)
```

Two regression tests added. `216 passed, 19 skipped, 1 xfailed` across `tests/mbb/test_mbb_torvik.py` + `tests/wbb/`; `codegen --check: all generated files current`.

Note the endpoint is healthy again today (both `2025_team_results.csv` and `2026_team_results.csv` return `text/csv`, HTTP 200), so this is hardening against a recurrence rather than a currently-broken source.

## Summary by Sourcery

Reject markup responses from Torvik CSV endpoints so transient outages produce clear source errors instead of downstream column failures.

Bug Fixes:
- Reject HTML and XML outage responses instead of parsing them as CSV data, allowing source-fetch failures to surface with an actionable error.
- Preserve the existing zero-row behavior for empty and other malformed non-HTML input.

Enhancements:
- Add focused regression coverage for markup detection, leading whitespace, UTF-8 BOMs, and CSV inputs beginning with angle brackets.

Documentation:
- Document the parser's HTML rejection behavior and retained malformed-input contract.

Tests:
- Add regression tests ensuring HTML outage pages raise clear errors without changing valid CSV parsing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of HTML error pages returned where Torvik CSV data is expected.
  * Raises a clear error instead of incorrectly parsing the HTML as CSV data.
  * Handles HTML responses with leading whitespace consistently.

* **Tests**
  * Added coverage for HTML outage pages and malformed responses.

### DIFF
--- a/sportsdataverse/mbb/torvik_parsers.py
+++ b/sportsdataverse/mbb/torvik_parsers.py
@@ -24,6 +24,12 @@ from sportsdataverse.dl_utils import underscore
 __all__ = ["parse_torvik_csv"]
 
 
+#: Document markers that mean "this body is markup, not a data file". Matched
+#: against the start of the payload, so a CSV cell containing "<html>" mid-file
+#: is unaffected.
+_HTML_MARKERS = ("<!doctype", "<html", "<?xml", "<head", "<body")
+
+
 def _clean_col(name: str) -> str:
     """janitor::make_clean_names-style cleaner for a single Torvik CSV header."""
     s = str(name).strip().replace("%", " percent ")
@@ -97,7 +103,10 @@ def parse_torvik_csv(payload: object, return_as_pandas: bool = False) -> Union[p
     # from here. The crosswalk builders wrap this in require_source(), whose
     # contract is exactly that a payload which will not render is a build
     # failure rather than a silently empty source.
-    if text.lstrip()[:1] == "<":
+    # Match real document markers, not a bare "<": a CSV whose first header
+    # cell is "<team>" is still a CSV, and a lone "<" belongs on the zero-row
+    # path below. A UTF-8 BOM ahead of the DOCTYPE is stripped first.
+    if text.lstrip("﻿").lstrip()[:200].lower().startswith(_HTML_MARKERS):
         snippet = " ".join(text.split())[:120]
         raise ValueError(
             "expected a barttorvik.com CSV data file but received an HTML "
--- a/sportsdataverse/mbb/torvik_parsers.py
+++ b/sportsdataverse/mbb/torvik_parsers.py
@@ -62,10 +62,17 @@ def parse_torvik_csv(payload: object, return_as_pandas: bool = False) -> Union[p
         de-duplicated column names; zero rows on empty/malformed input.
 
     Raises:
-        None: malformed input (non-text payload, header-only body, unterminated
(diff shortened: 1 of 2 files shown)


saiemgilani:
Good catch — fixed.

`text.lstrip()[:1] == "<"` was too broad on both counts you raised: a CSV whose first header cell is `<team>` is still a CSV, and a lone `<` belongs on the documented zero-row path, not the raising one. The guard now matches actual document markers (`<!doctype`, `<html`, `<?xml`, `<head`, `<body`) and strips a UTF-8 BOM first — a BOM ahead of the DOCTYPE would otherwise hide the very markup the guard exists to catch.

Both of your counterexamples are now tests:

```
  PASS  HTML doctype         want=raise got=raise  ValueError
  PASS  BOM + doctype        want=raise got=raise  ValueError
  PASS  <html> only          want=raise got=raise  ValueError
  PASS  xml                  want=raise got=raise  ValueError
  PASS  CSV with <team> hdr  want=parse got=parse  (1, 2) [team, conf]
  PASS  bare <               want=zero  got=zero   zero-row
  PASS  real CSV             want=parse got=parse  (1, 3) [rank, team, conf]
```

`11 passed` in `tests/mbb/test_mbb_torvik.py`; ruff clean.

PR #404, opened 2026-08-27
fix(transport): route the hand-written loaders through the classifier (#403)

### LINKED ISSUE #403: chore(transport): hand-written loaders bypass the release-read classifier
Split out of #397 / #402, which deliberately scoped itself to "change
`_read_release_parquet` once… no per-loader edits".

**The gap.** Two loader modules are hand-written and call `pl.read_parquet(url, ...)`
directly rather than going through `_codegen_runtime._read_release_parquet`, so they do
not inherit the failure classification added in #402:

- `sportsdataverse/nfl/nfl_loaders.py` — ~35 call sites (NFL is deliberately not in the
  codegen's `_GENERATED_LOADER_LEAGUES`)
- `sportsdataverse/cfb/cfb_loaders_extra.py` — 3 URLs (`CFB_BETTING_LINES_URL`,
  `CFB_TEAM_LOGO_URL`, `CFB_ROSTERS_CROSSWALK_URL`), each in a polars and a pandas branch

The 226 generated call sites now distinguish "asset is absent" (`NoDataError` → skip the
season) from "fetch failed" (`AssetFetchError` → surface it). These ~38 do not: a 403 or an
exhausted retry budget surfaces as whatever Arrow happens to raise, and any 404-handling
they do is per-site.

**Why it is worth doing.** The failure mode this guards against is silent — a rate-limited
season recorded as an empty one. NFL is the single largest loader family in the package, so
it is the worst place to lack the guard.

**Shape of the fix.** Not a drop-in swap: `_read_release_parquet` returns
`Optional[DataFrame]` (`None` on 404), whereas these sites currently propagate whatever the
reader raises. Either audit each site's expectations, or add a sibling helper that shares
the classification but keeps raise-on-missing semantics, so the change is mechanical and
behaviour-preserving.

**Note for whoever picks this up:** `tests/nfl/test_nfl_loaders_schema_drift.py` and
`tests/cfb/test_cfb_loaders_crosswalk.py` currently monkeypatch `pl.read_parquet` to stay
offline. They will need to patch the transport instead — and if that is done wrongly the
tests hit the live network and still pass, which is how a rewrite of them briefly pulled
94,602 real rows during #402. Assert the mock actually fired.

saiemgilani:
Closes #403. Follow-up to #402.

## The gap

#402 changed `_read_release_parquet` once and 226 generated call sites inherited
the new failure classification. Two hand-written modules fetch their own URLs and
inherited nothing:

| module | sites | why hand-written |
|---|---:|---|
| `sportsdataverse/nfl/nfl_loaders.py` | 38 | NFL is deliberately not in `_GENERATED_LOADER_LEAGUES` |
| `sportsdataverse/cfb/cfb_loaders_extra.py` | 6 | season-less crosswalk + logo/betting assets |

There, a 403, a rate limit and a missing asset were indistinguishable — all
surfaced as whatever Arrow happened to raise.

## Why not just give them the 404-safe helper

`_read_release_parquet` returns `None` for a missing asset, and `nfl_loaders.py`
has **zero** missing-season handling today — it raises. Swapping it in would have
silently converted a missing NFL season into a skipped one: a behaviour change,
and the wrong direction, since the whole point of #402 was to stop absent data
being confused with failed fetches.

## The split

One implementation, two entry points:

- **`_fetch_release_parquet`** — raises `NoDataError` when the asset is absent.
  Used by the hand-written loaders, so raise-on-missing is preserved; the error is
  now *typed and readable* rather than an opaque Arrow message, and it goes through
  the gateway's retry/pooling.
- **`_read_release_parquet`** — a thin wrapper converting only that one exception
  to `None`, for the generated loaders where a season gap is routine.

Only "absent" differs. `AssetFetchError` propagates from **both** — asserted
directly in `test_both_variants_surface_a_failed_fetch`, because a wrapper that
also swallowed a 403 would reintroduce exactly the silent-empty-season failure
#402 removed.

## The performance decision is still locked

The direct-read fast path is preserved in both entry points, and the
never-calls-the-transport guard now covers the raising variant too. The measured
+7% peak RSS / +75% wall regression from buffering cannot creep back in through
either door.

## Verification

- Live: a missing asset now raises `NoDataError` (previously an opaque reader
  error); `load_nfl_combine()` → (8968, 18) and `load_cfb_rosters_crosswalk()` →
  (37740, 14), both unchanged.
- Full suite: **5701 passed, 346 skipped, 10 xfailed**. Drift clean, ruff clean,
  mypy clean (400 files).

## Note

`CFB_ROSTER_URL` in `config.py` is now orphaned — `cfb_loaders.py` carries the same
URL as a literal. Left alone here deliberately: `cfb_loaders.py` is generated, so
deduping it means editing `releases.yaml`, which is a separate change and not worth
folding into a transport PR.

## Summary by Sourcery

Route hand-written parquet loaders through the shared classifier without changing missing-season semantics or direct-read performance.

Bug Fixes:
- Route hand-written NFL and CFB parquet loaders through shared transport handling so missing assets and failed fetches are surfaced with distinct, typed errors.

Enhancements:
- Provide separate raising and missing-season-tolerant parquet loader entry points while preserving the direct-read performance path and consistent failure classification.

Documentation:
- Document the distinct semantics of the raising and 404-safe parquet loader variants.

Tests:
- Add coverage for missing-asset behavior, failed-fetch propagation, fast-path transport avoidance, and metadata-refetch failures.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of missing sports data assets across NFL and college football loaders.
  * Missing assets are now safely skipped where appropriate, while failed or unauthorized requests consistently report an error instead of appearing as empty data.
  * Preserved existing output formats and data conversion behavior.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -786,9 +786,15 @@ recent exception when the retry budget is exhausted (instead of returning an
 unbound variable). Wrappers do NOT wrap the call in try/except — they trust
 `download()` to either return a usable `requests.Response` or raise.
 
-**Exception — remote columnar reads.** `_read_release_parquet()` in
-`_codegen_runtime.py` (backing 226 generated-loader call sites) hands the release
-URL straight to Arrow rather than buffering it through `download()`. This is
+**Exception — remote columnar reads.** `_fetch_release_parquet()` in
+`_codegen_runtime.py` hands the release URL straight to Arrow rather than
+buffering it through `download()`. Two variants share that one implementation:
+`_fetch_release_parquet` **raises** `NoDataError` when an asset is absent (used by
+the hand-written `nfl_loaders.py` and `cfb_loaders_extra.py`, where a missing
+season is an error), and `_read_release_parquet` returns **`None`** instead (used
+by the 226 generated call sites, where a season gap is routine). Only "absent"
+differs — a failed fetch raises `AssetFetchError` from both, so the wrapper can
+never soften a rate limit into an empty frame. This is
 deliberate and measured, not an oversight: Arrow overlaps the fetch with decoding
 and range-reads column chunks, while any fetch-then-parse path serializes the two
 and holds an extra copy of the compressed asset. On the 59 MB
--- a/sportsdataverse/_codegen_runtime.py
+++ b/sportsdataverse/_codegen_runtime.py
@@ -109,10 +109,10 @@ def cli_warn(msg: str) -> None:
     warnings.warn(msg, stacklevel=2)
 
 
-def _read_release_parquet(url: str) -> Optional[pl.DataFrame]:
-    """Read a release parquet; return ``None`` on 404 / missing asset (404-safe loaders).
+def _fetch_release_parquet(url: str) -> pl.DataFrame:
+    """Read a release parquet, RAISING when the asset is missing.
 
-    The read itself stays a direct Arrow fetch, and the bytes deliberately do NOT go
+    The read stays a direct Arrow fetch, and the bytes deliberately do NOT go
     through the HTTP gateway. Arrow overlaps the fetch with decoding and range-reads
     column chunks; buffering the asset first serializes the two and holds an extra copy
     of the compressed file. Measured on the 59 MB ``play_by_play_2025.parquet`` (best of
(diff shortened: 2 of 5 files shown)


saiemgilani:
Both review findings addressed — one fixed, one declined with reasoning.

**Fixed (`_codegen_runtime.py`, metadata fallback): real bug, good catch.**
`_read_parquet_stripped_metadata` issues its *own* request, separate from the read
that panicked, and parsed `resp.content` without checking the status. The first read
had reached the asset — that's why it panicked — but the refetch can independently
come back 403 or rate-limited, and that error body went straight to `pq.read_table`
and surfaced as a corrupt-parquet error. A failed fetch wearing the wrong clothes,
which is exactly the confusion this PR removes, one layer down. Now classified like
every other fetch: non-200 → `AssetFetchError`, 404 → `NoDataError` (keeping its
"absent" meaning through both entry points). Two regression tests added; the 403 one
fails with the status check removed. Fixed in f2cd69d.

**Declined (`nfl_loaders.py:382`, deprecated aliases): not accurate as stated.**
The aliases *do* dispatch to `load_nfl_nextgen_stats` / `load_nfl_pfr_advstats`
whenever `seasons` is passed. Only the `seasons is None` branch fetches directly,
and that's deliberate and commented — it preserves the original alias's "load every
season" behavior, which the unified entry point doesn't reproduce without an
explicit season list. The suggestion concedes this ("dispatch with an explicit
all-seasons value that preserves the current unfiltered result"), and synthesizing
that list would change what these loaders return.

It's also pre-existing: this PR only swapped the fetch call underneath, from
`pl.read_parquet` to `_fetch_release_parquet`. Rated 🔵 Trivial, and a behavior change
to deprecated loaders doesn't belong in a transport PR. Leaving it as-is rather than
silently skipping it.

PR #406, opened 2026-08-28
perf(ci): stop re-rendering the docs tree; split the offline and live test jobs

saiemgilani:
## The problem

The test job ran **60-100 minutes** and was routinely cancelled by the next merge before finishing — every recent run reads `cancelled`. In practice it verified nothing.

Profiling showed the cost was **not** spread across 6,000 tests. Four were **2,222s of a 2,755s offline run (81%)**, and almost all of it was the same work repeated: `_render_docs_all()` ran four to five times per session.

## The fix, measured

Only the determinism test needs a second render — that *is* its assertion. The other two just needed *a* corpus, so a session-scoped cache (`tests/codegen/conftest.py`) renders once and shares it.

| | before | after |
|---|---:|---:|
| `test_render_is_deterministic[_render_docs_all]` | 964.8s | 613.4s |
| `test_rendered_content_is_lf_only` | 416.9s | **<5s** |
| `test_generated_docs_tree_is_current` | 320.3s | **<5s** |
| `tests/codegen` (whole dir, serial) | 1043.6s | 678.9s |
| **offline job, exactly as CI runs it** | 60-100 min | **985s — 5,542 passed, 0 failed** |

Determinism still compares two **independent** invocations, so the property under test is unchanged; it just stops re-deriving input another test already produced. `_docs_stale()` gains an optional `rendered=` argument as the injection point, with default behaviour untouched.

## Why `--dist loadgroup` is load-bearing

A session fixture is scoped to a **worker**, not the run. Under `-n auto` these three tests can land on three workers and each re-renders, silently undoing the entire saving. They are marked `xdist_group("codegen_render")` and CI passes `--dist loadgroup` to pin them to one worker while the rest of the suite spreads out. This is correctness for the optimisation, not a tuning knob.

## CI-level dedupe

`codegen.yml` already runs **both** `tests/codegen/` and `generate.py --check`, on ubuntu, on the same triggers. The offline job now skips them — roughly 11 min per run for no extra signal. A bare `uv run pytest` locally still runs everything, and the pre-push hook still gates the drift check.

## Job split

- **offline** (ubuntu, parallel, no live network, no coverage) — fast enough to run on PRs again, restoring the merge-time signal removed on 2026-08-12.
- **live** (macOS, live APIs) — kept off the PR path.

macOS is **deliberately retained** rather than moved wholesale to ubuntu: it is currently the only platform under test, and the only job exercising the xgboost/`libomp` import path (which has its own dedicated step).

## Retry budget

`download()` hardcoded 15 retries at a 30s timeout with no override, so one unreachable host could park a single call for 15 x 30s plus backoff — right for a scraper that must not lose a game, wrong for a test suite.

`SDV_PY_HTTP_RETRIES` / `SDV_PY_HTTP_TIMEOUT` are now read **at call time** (so CI can set them after import). Defaults unchanged at 15/30, an explicit kwarg still wins, and a malformed or non-positive value falls back rather than raising — this runs on every request, so a typo must not take the process down. The live job sets 3/15. Four tests cover each rule.

## Gates

ruff clean · ruff format (1,286 files) · mypy clean (400 files) · codegen drift current · `tests/codegen` 168 passed under `-n auto --dist loadgroup` · offline job 5,542 passed / 0 failed.

## Note for reviewers

`uv.lock` on main is stale — `pyproject.toml` says `0.1.2`, the lock says `0.1.1`. I deliberately kept that out of this PR. It is **not** breaking CI (`uv sync --frozen` tolerates a local project-version mismatch; verified), but it wants a separate deliberate re-lock.

## Summary by Sourcery

Restore practical CI feedback by eliminating redundant code-generation work, splitting offline and live test execution, and bounding live API retry budgets.

Bug Fixes:
- Allow HTTP retry counts and timeouts to be configured through environment variables while preserving existing defaults and explicit argument precedence.

Enhancements:
- Cache shared code-generation renders to avoid redundant documentation-tree generation while preserving determinism checks.
- Allow documentation staleness checks to consume an already-rendered corpus.

CI:
- Split CI into a fast parallel offline test job that runs on pull requests and a macOS live-API job that remains off the pull-request path.
- Use xdist load-group scheduling for code-generation tests and remove duplicate codegen checks from the offline job.
- Bound live-job HTTP retries and timeouts through CI environment settings.

Tests:
- Add coverage for environment-configured HTTP retry behavior and argument precedence.

## Summary by CodeRabbit

* **New Features**
  * HTTP request retry counts and timeouts can now be configured through environment variables.
  * Explicit request settings, including disabling timeouts, take priority over environment defaults.
  * Invalid or non-positive configuration values safely fall back to standard defaults.

* **Bug Fixes**
  * Improved validation reliability and reduced unnecessary repeated generated-content rendering.

* **Tests**
  * Added coverage for configurable request behavior, fallback handling, and deterministic generated output.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,56 +1,49 @@
-# Run the full test suite (live API tests included) on push to main and
-# workflow_dispatch. Live tests are gated by SDV_PY_LIVE_TESTS=1 and run by
-# default; the workflow_dispatch input can flip them off (live_tests:
-# "false") for runs where third-party API churn would produce
-# false-negative failures.
+# Test suite, split into a fast offline job and a slower live-API job.
 #
-# SCOPE (deliberately narrowed 2026-08-12, operator decision): a SINGLE job —
-# macos-latest on the canonical interpreter — running ONLY on main. Two
-# consequences to know about, because neither is caught anywhere else:
-#   - Python 3.9 (the declared floor) and Linux/Windows are no longer
-#     exercised. CONTRIBUTING.md promises the floor is covered; it is not.
-#     A 3.9-incompatible construct or a platform-specific regression now
-#     ships and surfaces at release time.
-#   - PRs no longer run tests. Merge-time signal is the codegen drift gate
-#     plus whatever ran locally, so `uv run pytest` before pushing is now
-#     load-bearing rather than a convenience.
-# Restore by re-adding the matrix `include:` block and the `pull_request:`
-# trigger below.
+# WHY THE SPLIT (2026-08-27): the single combined job ran 60-100 minutes and was
+# routinely cancelled by the next merge before it finished, so in practice the
+# suite verified nothing. Profiling found the cost was NOT spread across 6,000
+# tests -- four codegen tests were 2,222s of a 2,755s offline run (81%), almost
+# all of it the SAME doc render repeated four to five times. That is fixed in
+# tests/codegen/conftest.py; the jobs below carry the rest.
 #
-# Uses `uv sync --all-extras --all-groups` against the committed uv.lock for
-# reproducible installs, then `uv run --frozen pytest` for execution. See
-# CONTRIBUTING.md.
+#   offline  ubuntu, parallel, no live network. Fast enough to run on PRs, which
+#            restores the pre-2026-08-12 merge-time signal.
+#   live     macOS, live APIs. Keeps the platform the narrowing kept -- it is the
+#            only job that exercises the xgboost/libomp import path -- and keeps
+#            the live coverage, off the critical path.
+#
+# SCOPE (still narrowed, unchanged by this split): Python 3.9 (the declared floor)
+# and Windows are NOT exercised. CONTRIBUTING.md promises the floor is covered; it
+# is not. Restore by re-adding a matrix `include:` block.
+#
+# Uses `uv sync --all-extras --all-groups` against the committed uv.lock.
 name: tests
+
 on:
   push:
     branches: [main]
+  pull_request:
   workflow_dispatch:
     inputs:
       live_tests:
         description: "Set SDV_PY_LIVE_TESTS=1 to include live API tests"
         required: false
         default: "true"
-# Cancel in-flight runs on the same ref when a new push comes in.
+
+# Cancel superseded runs on the same ref. On main this used to mean each merge
+# killed the previous run mid-flight and the suite never completed on any commit;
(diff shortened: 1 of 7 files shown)


saiemgilani:
Both findings addressed in 8c5f32c — the second was a real defect I'd introduced.

**`timeout=None` semantics (functional correctness): fixed.** Correct catch.
`None` is meaningful to `requests` (no timeout, wait forever), and defaulting the
parameter to `None` quietly reinterpreted it as "use 30s".

Verified the blast radius before fixing: no caller in this repo is affected. Both
variable call sites are typed `int` with a default of 30 (`hockeytech/_client.py`,
`nfl/ep_wp.py`), and `_HTTP.send_api_request`'s `timeout=None` goes to
`requests.get` directly, never through `download`. But `download` is public API,
so an external caller passing `None` deliberately would have silently lost their
"no timeout". A `_UNSET` sentinel now keeps "omitted" and "explicitly None" apart:
only an omitted argument consults the environment, and an explicit `None` is
forwarded unchanged. Covered by
`test_explicit_none_timeout_still_means_no_timeout`, which asserts both directions.

**Docstring contract: fixed.** Added a `Raises:` block (`NoDataError` on a 404 or
ESPN's 200-with-`code:404` body, not retried; `RequestException` when the retry
budget is exhausted) and documented `SDV_PY_HTTP_TIMEOUT` / `SDV_PY_HTTP_RETRIES`,
the explicit-argument > environment > default precedence, call-time rather than
import-time resolution, and that a malformed or non-positive env value falls back
instead of raising.

Worth noting for anyone reading the diff: my first attempt at that docstring block
landed it in `_env_int` instead of `download` (the insert anchored on the file's
first `Returns:`). Caught and moved; `_env_int`'s docstring is clean.

saiemgilani:
Taken in 5176df9, with one correction to the premise.

**Not a regression.** The finding says "the previous default was `None`" — on main
the default is `15`, and `max(int(num_retries), 0)` rejects `None` there too, so
`num_retries=None` raised `TypeError` before this PR as well. Verified both ways.
It was `None` only in this branch's intermediate commit, which never shipped.

**Taking the fix regardless**, because the reasoning underneath it is right: unlike
`timeout`, `None` has no meaningful retry semantics — there is no "infinite
retries" reading — so treating it as omitted only widens what is accepted and takes
nothing away. The asymmetry with `timeout` is genuinely surprising, and the guard is
one line. Documented on the parameter and covered by
`test_none_num_retries_is_treated_as_omitted`.

PR #408, opened 2026-08-29
fix(cfb): repair the EP/WP inputs — field position, possession perspective, and era vocabulary

saiemgilani:
Ten commits repairing the expected-points and win-probability inputs in the CFB
play-by-play parser. What began as one reported EPA anomaly turned into an audit of the
class it belonged to, then of every era the parser covers, then of the branch's own
shipped work. Each fix is measured against real seasons and backtested; none is inferred.

**Every figure here is measured on release data — `load_cfb_pbp`, 2004–2025 — not on
fixtures or estimates.**

---

### Win probability was reported for the wrong team on every possession change

`wp_after` is stated in the **start**-possession team's perspective. The home/away split
keyed off `end.pos_team.id` — the same team on most plays, the **other** team on every
possession change — so `home_wp_after` and `away_wp_after` came out complemented on
exactly those rows.

Measured against the identity that home WP at a play's end equals home WP at the next
play's start, over published 2023–24:

| group | n | end-map | start-map |
|---|---|---|---|
| possession unchanged | 289,741 | 0.0061 | 0.0061 *(same team either way)* |
| **changed + non-scoring** | **24,332** | **0.6797** | **0.0099** |
| changed + scoring | 654 | 0.6131 | 0.1475 |
| changed + onside kick | 11 | 0.6824 | 0.1271 |
| end of half | 1,859 | 0.0929 | 0.0072 |

**17,125 of the 24,332** possession-change non-scoring rows were off by more than 0.5.
`wpa` never revealed it: that is `wp_after − wp_before` and both sit in the same
perspective, so the error cancels there and surfaces only in the exports.

### Win probability inverted by a dead-ball foul

Separately, `wp_after` is derived from `lead_wp_before` via `1 - lead_wp_before` — a
perspective conversion valid only when the next play is genuinely the other team's
possession.

| rows reaching the branch, 2022–24 | n | |
|---|---|---|
| `lead_pos_team == end.pos_team.id` | 35,501 (97.0%) | flip correct — punts, turnovers |
| `lead_pos_team == start.pos_team.id` | **867 (2.4%)** | **flip returns the complement** |

**515 of the 867 published `|WPA| > 0.5`**, worst `wp 0.002 → 0.998`. Subset mean WPA was
**+0.145** where WPA must centre on zero; after the guard, 6 rows above 0.5 and a mean of
+0.011. A re-kick kickoff penalty is the clearest generator (30.6% flip rate vs 0.25%),
but punts (390) and missed field goals (94) dominate by count. 2025 adds another 872 rows.

---

## Expected points

**Mirrored `end.yardsToEndzone`.** A 2-yard rush scored **+6.45 EPA** on
[game 401868040](https://gameonpaper.com/game/401868040) because a pass two plays earlier
carried `end.yardLine=5` with `end.yardsToEndzone=95`. There is no same-row test that
could catch it — `yardsToEndzone == yardLine` on 3,848 non-kick plays and equals its
complement on 3,859 — so the repair keys on the next play's start being the exact
complement, with possession unchanged across the pair. The 34-yard gain goes −2.37 → +3.14
and the following penalty 6.45 → 0.94.

**Penalty-only kickoff rows.** ESPN writes the penalty's *enforcement spot* — the kicking
team's own territory — into `end.yardsToEndzone` while `pos_team` is the receiving team, so
a 15-yard flag **against** the receiving team published as **+5.52 EPA for it**. 83 such
rows. Split by whether the ball is re-kicked: a re-kick makes the row a no-play ending at
the touchback yardline (42 rows, averaging +0.78 EPA and reaching +3.92); no re-kick means
the next real play is the only record of where the ball went (24 rows, disagreeing with the
stored value 17 times by a mean of 14 yards).

**`EP_between` folded across a score.** The term measures *unexplained* discontinuity, and
after a score there is none — the points explain it, and `lag_EP_end` is a realized 7.00.
26 scrimmage penalties across 2005–2025 fall where the fold applies after a score and
**14 (54%) published `|EPA| > 4`**; zeroing takes that to 0 and the maximum from 8.39 to
3.25.

**Kick-return end states where ESPN contradicts its own text.** The raw payload for
401762869 p122 carries `end.yardLine: 8` and `end.yardsToEndzone: 8` while its text ends
`"to EMU08"` — the possessing team on their own 8, which is 92 to go.

The text alone is **not** sufficient, and that is the design. Where text and field disagree
the field is usually right — 313 of 326 adjudicable cases in 2005, 88 of 94 in 2018 —
because side resolution fails on older abbreviations. Only 2025 inverts. So the gate
requires **two independent sources against the field**: the text's final spot, and the next
real play starting exactly there with possession unchanged. Class scope was adjudicated
individually: kickoffs, punts and interceptions are in; **fumble returns are out**, because
there the field beats the text 27 to 9.

---

## Era coverage

ESPN changed the CFB text template twice, and the changeovers are total rather than gradual.
Three fixes on this branch were tuned to one vocabulary and mishandled another — all caught
by re-auditing, not by review.

| feature | 2004–2013 | 2014–2024 | 2025 |
|---|---|---|---|
| kick distance | `for N yards` (98%) | `for N yds` (97–99%) | `kickoff N yards`, no "for" — only **57%** carry the 2014 form |
| out of bounds | `out-of-bounds` | `out of bounds` | |
| field goal | `24 yard field goal GOOD` | `23 yd FG GOOD` | both forms |
| extra point | `extra point GOOD` (5,942 in 2013) | folded into `(Kicker Kick)` — **0** from 2018 | |
| `penalty_text` | `10 yard holding ACCEPTED NO PLAY.` | `, Illegal Block (10 Yards) to the TexAM 10` | |
| enforcement spot | **absent** | present (9,856 of 10,355) | |
| penalty yardage | **NET** spot change, reaches 29–34 | nominal, tops ~25 | |

Consequences fixed: the penalty-yardage bound is now era-aware (a flat 25 discarded 164
legitimate 2007 rows); part (d)'s kick-outcome gate recognises all three vocabularies
(anchored on the word *kickoff*, because a bare "N Yards" is also how the penalty yardage
reads); part (c)'s kick exclusion covers `FG` and `(Kicker Kick)`.

---

## Names

**`_FORMATION_PREFIX` truncated real surnames — released in v0.1.3.** Its trailing `[\s\-]*`
matches the empty string, on the stated assumption that "no real surname contains huddle or
shotgun". Huddleston does: `"Rakeem Cato sacked by Parrish Huddleston"` became **`"ston"`**.
Published assets predate v0.1.3 and are clean, but a rebuild would have carried it.

**Capture-window truncation.** The name extractions use fixed-width windows, and when ESPN
appends the next play's text the window stores 29 characters of it as a name. Every
malformed value sits at its window's ceiling. `fumble_recovered_player_name` (126 rows,
8.4% of published 2023) was already cured on main and wants only a re-publish;
`fg_block_player_name` was still live and is fixed here.

**2025's vendor template** opens each play with the game clock, so 55,135 published 2025
name values hold play-text debris — 51,400 `athlete_name`, 3,682 `kickoff_player_name` of
the form `"(15:00) #36 T.Morrison"`.

---

## Penalty effect, computed directly (commits 11–15)

`EPA_penalty` measures the whole play. **`EPA_penalty_direct`** isolates the flag:
`EP(actual end) − EP(counterfactual end)`, where the counterfactual is where the play would
have ended with no flag — `start − play_yards`, scored through the same EP model from the same
frame. Additive column; `EPA_penalty` is unchanged.

That is deliberately not `end + penalty_yards_net`. The two agree on a tack-on and disagree on
a spot-of-foul enforcement: a 22-yard run with offensive holding ends at `start + 3`, and the
flag's cost is the 22 yards *plus* 10. Worked cases: an offensive holding on a 6-yard run is
**−1.00** where the play-level value was −0.44; a defensive holding tacked onto a 4-yard run
is **+0.63** where `EPA_penalty` credited the whole play at +1.75.

Scope is narrow on purpose — rush / completion / incompletion / sack, accepted, on a play that
stands, not scoring / turnover / end-of-half / fourth-down non-conversion — and **null outside
it**. Declined and offsetting flags are zero (they had been inheriting the play's swing on ~97%
of rows: a declined defensive holding published +3.50). No-play and standalone rows use
`EP_end − EP_start`, since the start state *is* their counterfactual.

**Its inputs were verified, not assumed.** `end.yardsToEndzone` *includes* the enforcement on a
stands row (78% of rushes equal `yds_rushed ± penalty`); `statYardage` is the net; the
published `end.*` columns are proven to be the model's inputs (instrumented `DMatrix` calls,
181/181 rows) while `down_*_end` is reset independently of `end.down`, so the counterfactual
builds its own flags. The sign of the penalty's yardage — needed for nothing here, but for
`penalty_yards_net` — resolves only when two of three independent signals agree
(`penalty_side`: 59.5% coverage at 99.62% on 19,608 labelled rows).

**The counterfactual's sign against `penalty_side` became a detector**, and three input defects
fell out:

| defect | scale | fix |
|---|---|---|
| `yds_sacked` grabbed the **yardline** in 2004–07 text (`"at the ECaro 48 for a loss of 8"` → −48) | 2,227/2,283 sacks in 2005, 2,395/2,424 in 2006, 2,806/2,843 in 2007 | stated loss wins in every era form; a sack stating no loss is null |
| first-down signal fired on plays that **stand** (the play converted, not the flag) | offensive Illegal Block resolved `def` | L1 requires a no-play row |
| part (c) excluded penalty rows from the mirror repair | **229 of 2,045** 2025 scrimmage-penalty rows stored as the complement (`"to NEV1"` → 90) | widened; the exact-complement-with-possession-unchanged signature is the mirror regardless |

## Verification

**362 games, 2004–2026** — 16 per season plus 10 completed 2026 games fetched live (no release
asset exists yet) — run before and after on identical games: **65,386 plays, 501 shared
columns**. Every changed cell is accounted for: `yds_sacked` (169 rows), `end.yardsToEndzone`
(12), `penalty_side`/`net` (303), and their propagation — `lag_EP_end` and `EP_between` on
exactly the row after each mirror repair, `prog_drive_*` within the same drives. **Zero
unexplained.** The last-added optional-column guard is proven a no-op on real frames (three
games re-run, 0 of 501 columns differ).

Earlier gates on this branch: the 13-game era set (2004–2025), the part (e) backtest (53
games, 17 changed / 17 targeted / 0 unintended), and the rejected widening to punts and
interceptions (26 rows moved, EPA distribution worse — recorded so it is not retried).

Every regression test confirmed failing against the prior code. Full `tests/cfb`: **793 passed,
55 skipped, 2 xfailed**. `ruff format` and `ruff check` clean.

## Review findings

CodeRabbit's MAJOR finding claimed part (e)'s branch introduced a perspective mismatch. It did
not — that branch and the pre-existing `1 - lead_wp_before` both yield start-perspective — but
chasing it found the 24,332-row home/away defect above, the largest thing in this PR. Also
taken: part (d.ii) shares (d.i)'s administrative-row lookahead; midfield and team-qualified
spots compete on position; `_penalty_spot` in the drop list. Not taken: the fixture-guard
suggestion.

## Not addressed

Punt / interception / fumble return end states (the field beats the text on fumbles; on the
others the text beats the field yet the widening still worsened EPA — needs its own
investigation); a re-publish of 2023–25 assets, which several of these fixes warrant; and the
NFL twin of the WP construct in `ep_wp.py:2185`, unmeasured because the published NFL release
carries the nflfastR schema.

## Summary by Sourcery

Repair CFB play-by-play probability, expected-points, penalty, attribution, and player-name parsing across historical and current ESPN data.

New Features:
- Add direct penalty EPA values that isolate the flag’s impact from the underlying play using a scoped counterfactual end state.
- Expose penalty enforcement spots, penalty-side resolution, signed net yardage, and related attribution fields.

Bug Fixes:
- Correct win-probability perspective handling across possession changes, re-kicks, game-ending plays, and home/away exports.
- Repair malformed expected-points inputs caused by mirrored end-state distances, kickoff penalty enforcement spots, kick-return positions, and score-transition discontinuities.
- Correct sack yardage, parenthesized penalty yardage, penalty-side inference, and era-specific penalty magnitude handling.
- Prevent malformed player names from formation prefixes, game-clock and jersey tokens, and blocked-field-goal capture-window debris.

Enhancements:
- Expand parsing and validation across historical, modern, and 2025 CFB text vocabularies while limiting ambiguous return-state repairs to adjudicated kickoff cases.

Tests:
- Add fixture-backed regression coverage for corrected WP, EP, penalty, return-state, sack-yardage, name-parsing, and era-vocabulary behavior.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected win-probability perspective handling during re-kicks, punts, missed field goals, and similar possession changes.
  * Prevented kickoff penalty enforcement spots from being treated as field position when no kick or return occurred.
  * Improved expected-points and field-position calculations for penalty plays.
  * Repaired mirrored end-of-play yardline values and improved penalty-yardage validation.

* **Tests**
  * Added coverage for kickoff penalties, re-kicks, mirrored yardlines, kicking plays, enforcement spots, and implausible penalty-yardage values.


saiemgilani:
## Follow-up: a second, related case this PR deliberately does NOT fix

Reviewing the sequence around the repaired play surfaced the same corruption on a **kickoff**, which this PR's predicate excludes. Recording it rather than widening the rule.

Game 401868040, play #107:

```
text : "#17 S.Braidwood kickoff 64 yards to the VIL01 #4 J.Mace return 19 yards to the VIL20 ... PENALTY"
start: team=2729 (kicking)   y2ez=65  yardLine=65
end  : team=222  (receiving) y2ez=10  yardLine=10     <- mirrored
next : team=222              y2ez=90  yardLine=10
=> EPA +4.13 on a kickoff return
```

The return reached Villanova's own 20 (80 to go), a penalty then pushed them to their own 10, and the next play starts at 90 — so the next-start is right and `end.y2ez=10` is wrong. Same mirror, different play class.

It is excluded because possession changes *within* a kickoff (`start.team` 2729 vs `end.team` 222), so the `start.team == end.team` guard rejects it.

### Why I am not widening the predicate here

Relaxing to `end.team == next.start.team` and anchoring on a stable `yardLine` grows the corpus hits from 54 to 122 (66 plain, 39 kick, 17 penalty). That sweep includes rows like:

```
Timeout Ohio State, clock 12:45   y2ez 73 -> 27, yardLine 73 on both sides
```

A timeout cannot move the ball, so one of those two numbers is wrong — but **nothing available establishes which**. `yardsToEndzone` equals `yardLine` on 3,848 non-kick plays in the corpus and equals its complement on 3,859, so `yardLine` cannot arbitrate. Trusting the next-start universally would be an assumption, not a finding, and on kickoffs it would rewrite EPA on every game.

The narrow rule in this PR is verified end to end: 54 rows, 0 possession mismatches, 0 kicks, and the reported play goes 6.45 -> 0.94.

Widening it needs an independent authority for which side is correct — most likely reconciling against the drive's start/end field positions, or cfbfastR as the reference implementation. That is its own change with its own validation, not a rider on this one.

PR #426, opened 2026-09-02
fix(cfb): field-position flags measured the wrong end of the field

saiemgilani:
## The bug

`rz_play`, `scoring_opp` and `goal_to_go` were computed from `start.yardLine` — ESPN's absolute 0-100 field
position — rather than the distance to the opponent's goal line. `yardLine` does not flip with possession, so
a test like `<= 20` selects plays pinned against one particular end zone: the red zone for the team driving
that way, and its own shadow for the team driving the other.

## Evidence

Scrimmage plays only, three 2026 games:

| game | really inside the 20 | `rz_play` said | really goal-to-go | `goal_to_go` said |
|---|---|---|---|---|
| 401864570 | 24 | 14 | 10 | **0** |
| 401856766 | 24 | 14 | 5 | 6 |
| 401864494 | 34 | 16 | 18 | 9 |

On `401864570` `goal_to_go` is false on all **8** plays ESPN itself labels `& Goal` (`1st & Goal at NMSU 9`,
`3rd & Goal at NMSU 4`, …) and true on no play in the game.

## The fix

All three now read yards-to-goal:

- `rz_play` — `start.yardsToEndzone <= 20`
- `scoring_opp` — `start.yardsToEndzone <= 40`
- `goal_to_go` — `start.distance >= start.yardsToEndzone`, which is what goal-to-go means

810 CFB tests pass. No fixture depended on the previous behaviour.

## Blast radius

Anything consuming these columns was wrong on roughly half the real red-zone snaps — the `EPA_success_rz`
family in `create_box_score`, and Game on Paper's red-zone and scoring-opportunity panels downstream of it.

## Summary by Sourcery

Measure CFB field-position flags from the opponent’s goal line so downstream red-zone and scoring-opportunity metrics classify plays correctly.

Bug Fixes:
- Correct field-position flags to measure distance to the opponent’s goal line, fixing red-zone, scoring-opportunity, and goal-to-go classification across possessions.

Tests:
- Verify the corrected CFB field-position flags with the existing 810-test suite.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected field-position indicators in college football play-by-play data to consistently measure distance from the opponent’s goal line.
  * Improved identification of red-zone plays, scoring opportunities, and goal-to-go situations across different field positions.

### DIFF
--- a/sportsdataverse/cfb/cfb_pbp.py
+++ b/sportsdataverse/cfb/cfb_pbp.py
@@ -6574,7 +6574,10 @@ class CFBPlayProcess(object):
                 # 14 of the 24 real red-zone snaps.
                 rz_play=pl.when(pl.col("start.yardsToEndzone") <= 20).then(True).otherwise(False),
                 under_2=pl.when(pl.col("start.TimeSecsRem") <= 120).then(True).otherwise(False),
-                goal_to_go=pl.when(pl.col("start.distance") >= pl.col("start.yardsToEndzone"))
+                # goal-to-go is an equality, not a threshold: the line to gain IS the goal line, so
+                # the distance needed equals the yards remaining. `>=` would also swallow a feed row
+                # where distance exceeds yards-to-goal, which is not a real down and distance.
+                goal_to_go=pl.when(pl.col("start.distance") == pl.col("start.yardsToEndzone"))
                 .then(True)
                 .otherwise(False),
                 scoring_opp=pl.when(pl.col("start.yardsToEndzone") <= 40).then(True).otherwise(False),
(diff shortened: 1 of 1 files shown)


saiemgilani:
Corrected `goal_to_go` to an equality per review: the line to gain **is** the goal line, so the distance needed equals the yards remaining.

Validated against ESPN's own down-and-distance text across four 2026 games — `start.distance == start.yardsToEndzone` reproduces the `& Goal` label exactly, with no false positives or negatives:

| game | `distance == yardsToEndzone` | ESPN text says `& Goal` |
|---|---|---|
| 401864570 | 10 | 10 |
| 401856766 | 5 | 5 |
| 401864494 | 18 | 18 |
| 401858202 | 5 | 5 |

`distance > yardsToEndzone` occurs on **0** plays in all four, so the previous `>=` agreed on this data — but such a row is not a real down and distance, and `>=` would swallow it silently rather than leaving it out.

810 CFB tests still pass.

PR #451, opened 2026-09-03
feat: wrap the remaining hoopR/wehoop basketball sources (KenPom, Her Hoop Stats, Basketball-Reference, RealGM, salary/draft/injury)

saiemgilani:
## What

Wraps the four remaining basketball sources that existed only in hoopR/wehoop, plus the shared layer they need.

| Source | Python | Covers |
|---|---|---|
| KenPom | 30 codegen wrappers (`kenpom_*`) | hoopR's 36 `kp_*()` |
| Her Hoop Stats | 3 fns + login helpers | wehoop's 3 `hhs_*()` |
| Basketball-Reference | 9 fns behind `league=` | 9 hoopR + 3 wehoop |
| Spotrac / HoopsHype / NBADraft / RotoWire | 4 fns | 4 hoopR |
| RealGM | 17 fns | 17 hoopR |

## The shared layer

`_html_tables.py` replaces the ~44 hardcoded `header_cols` vectors hoopR carries. Those exist because `rvest::html_table()` mangles a grouped `<thead>`; `pandas.read_html` builds a real MultiIndex, so two rules reproduce the intended names generically and a new upstream column widens the frame instead of shifting every column.

`_subscription_http.py` handles login, proxy and session caching for the two paywalled sources. It **verifies the login actually took** — both sites return HTTP 200 with the login page on a rejected login, so anything that doesn't check scrapes free-tier tables and calls them subscriber data. The R packages don't check.

## Live-verified

KenPom and Her Hoop Stats were run against real subscriptions on 2026-09-02, which found seven bugs no offline fixture could:

- KenPom repeats its header as **10 separate `<thead>` elements**, so pandas reports a 20-level MultiIndex whose labels alternate — collapsing only adjacent duplicates repeated one name ten times inside a single column name.
- Header rows also land in `<tbody>`: 382 rows for 364 teams.
- The NCAA seed is glued to the team label (`"Duke 1"`), silently breaking joins for exactly the 68 tournament teams.
- Everything is text with `+`-signed values.
- Her Hoop Stats embeds a name-pronunciation `<audio>` in the player cell, and `read_html` concatenates descendant text, so names arrived as `"Te-Hina Paopao  This HTML5 audio…"`.
- wehoop pairs team links to rows by position, which shifts every later URL if any row gains a second anchor; this walks rows instead.
- `team_link` was missing entirely, making the other two functions uncallable.

Six regression tests pin these against trimmed real captures. Fixtures carry no credentials or account chrome (verified).

## Notes

- **RealGM needs a residential IP.** Its Cloudflare challenge is designed to fail on datacenter addresses, so it must not be scheduled in CI or on the droplet. The browser hop is Playwright (already an optional extra) and fully injectable, so tests need no browser.
- Basketball-Reference deliberately does **not** use the shared table reader — Sports Reference's real column names live in `data-stat` attributes, which is why it gets its own parser.
- Reference docs for all of it live in `sdv-internal-refs` (`kenpom/`, `herhoopstats/`, `rotowire/`, plus port-status sections).

## Also in here

`fix(mbb)`: the NCAA fetch docstring claimed new-headless is required because it renders through the real GPU/ANGLE path, that SwiftShader is "a textbook Akamai tell", and that a GPU-less host "is re-detected — run on a real-GPU host." That is measurably wrong: on a virtio-GPU droplet Chrome falls back to SwiftShader and bm-verify clears normally — 342 bundles, zero challenge failures, zero bans (`ncaa-mbb-hoops-raw/docs/SCRAPING_NOTES.md`, 2026-08-02). The WebGL renderer string is not part of the check. Docstrings only; no behaviour change, and the two tells the docstring already identified correctly are untouched.

## Verification

`ruff` clean, `mypy` clean on the new modules, codegen drift gate clean, `uv.lock` untouched. 2,163 tests pass across `tests/nba`, `tests/mbb`, `tests/wbb` and the subscription suites. Reference docs for all of it are in the companion `sdv-internal-refs` PR.

## Summary by CodeRabbit

- **New Features**
  - Added authenticated KenPom college basketball data access, covering ratings, teams, players, games, conferences, officials, trends, and more.
  - Added Her Hoop Stats women’s basketball access for teams, rosters, and statistics.
  - Added NBA data retrieval from Basketball-Reference, RealGM, HoopsHype, NBADraft.net, RotoWire, and Spotrac.
  - Results support Polars or pandas formats, with improved handling of unavailable or incomplete source data.

- **Documentation**
  - Expanded reference documentation and updated available-function counts across MBB, NBA, and WBB.


saiemgilani:
All six addressed in `aa828b49`. Each was verified against the code before changing anything; all six were real.

**SSRF (CWE-918), `wbb/herhoopstats.py`** — correct and the most serious of the six. `team_link` is a public argument that steers an *authenticated* request, so an absolute URL sent the subscriber's logged-in session anywhere. Now restricted to relative paths and `https` on herhoopstats.com, with the `team_link` contract updated. Verified behaviourally rather than by eye:

```
ALLOW  /stats/ncaa/team/12345/2024/     refuse https://evil.example/steal
ALLOW  https://herhoopstats.com/stats/x refuse http://herhoopstats.com/x
ALLOW  https://www.herhoopstats.com/y   refuse //evil.example/x
                                        refuse https://herhoopstats.com.evil.example/x
                                        refuse file:///etc/passwd
```

**Both races** — real. `bref._throttle` read/slept/stamped unserialized, so two threads compute the same `wait` and fire together, silently halving the pacing. `realgm._playwright_html` shares one page per process, so an interleaved `goto()` makes an earlier call return HTML from the wrong endpoint — a wrong-data bug, not a slow one. Both now hold a module lock across the full sequence.

**RealGM TTL** — real. The TTL is documented as *idle* time but started before navigation, so a slow Cloudflare challenge could expire the session it had just built. It now restarts once the fetch succeeds.

**The dict annotation** — real, and fixed in the **generator** rather than by hand across 30 wrappers. `api_module.py.jinja` hardcoded `-> Union[pl.DataFrame, pd.DataFrame, …]` regardless of what the parser returns, so this would recur for any dict-returning parser. Added a `parsed_types` field alongside the existing `raw_types` — same YAML plumbing, defaulting to the current single-frame union, so **every other flat-API stem is byte-identical** (checked: NHL/MLB still emit `pl.DataFrame, pd.DataFrame, Dict`). KenPom's YAML now declares the dict shape and all 30 wrappers regenerated. Worth noting its `parsed_doc` already said *"a dict of polars DataFrames"* — only the annotation was lying.

**The `y=2025` example** — real; the parameter is `year`, so the documented example raised `TypeError`.

Verified: 2,314 passed, 77 skipped, 1 xfailed; codegen regenerated and the docs tree is current (the drift gate initially caught a stale `wbb/reference/additional.md` from the docstring edit — regenerated).

PR #456, opened 2026-09-03
feat(cfbd): wrap the College Football Data API — 58 endpoints

saiemgilani:
CFBD is the largest single source the Python side never reached — cfbfastR gets there through 78 R functions, sdv-py through none.

**The backlog listed this as blocked on a "free key". That was stale.** The key has been in `.Renviron` all along and the OpenAPI spec has been sitting in `sdv-swagger`; the only thing missing was the work.

## Generated from the spec, not transcribed

`tools/codegen/gen_cfbd.py` emits `endpoints/cfbd.yaml` — **58 endpoints, 218 query params** — so the surface tracks the published contract. **17 endpoints correctly carry required parameters**, because the generator honours the spec's `required` flag and the template already emits a no-default argument for it.

Two traps in the source file, both documented in the generator:
- `cfbd-swagger.json` is a **swagger-UI config dump**, not a bare OpenAPI document — the spec is nested under `swaggerDoc`, and the file ends with a stray `;` that makes `json.load` fail outright.
- It resolves its own path by **walking up** rather than a fixed `parents[N]` hop, which breaks inside a git worktree.

## Auth resolves at call time

`cfbd_runtime._get` tries explicit `headers`, then the environment, then `~/.Renviron`. That last hop matters here specifically: **the SportsDataverse keys live in `.Renviron`, which R reads at startup and Python does not** — so a key that is plainly "set" for the R packages is invisible to `os.environ`. Resolving lazily keeps `import sportsdataverse` working with no key present, and keeps the key out of tracebacks.

One parser covers all 58 routes because CFBD is uniform (a flat array of records), with one level of nesting flattened and lists stringified.

## Verified live against the real API — 8/8 sampled wrappers

| wrapper | rows | cols |
|---|---:|---:|
| `cfbd_venues` | 852 | 14 |
| `cfbd_teams` | 678 | 26 |
| `cfbd_conferences` | 256 | 6 |
| `cfbd_games` | 231 | 42 |
| `cfbd_recruiting_teams` | 194 | 4 |
| `cfbd_ratings_sp` | 135 | 27 |
| `cfbd_talent` | 134 | 3 |
| `cfbd_rankings` | 1 | 4 |

One bug caught by that live check rather than by review: the flat-API contract passes a **fully-qualified URL** to the runtime, and my first `_get` treated the argument as a path and prepended the host — producing `api.collegefootballdata.comhttps`. Fixed to accept either form.

Follow-up (not in this PR): returns-schemas under `schemas/native/cfbd/` are referenced but not yet authored, and CBBD is the same shape once this lands.

## Summary by CodeRabbit

- **New Features**
  - Added access to 58 College Football Data API endpoints through the CFB module.
  - Supports authenticated requests using an API key supplied directly or through environment configuration.
  - API responses can be returned as tidy Polars or pandas data frames, or as raw JSON.
  - Added handling for nested response fields and empty results.

- **Documentation**
  - Added comprehensive endpoint reference documentation, usage examples, and Python–R parity links.


saiemgilani:
Closing — I built this against an item that had already been **dropped**.

The stocktake records it as *"DROPPED 2026-09-02 by Saiem: 'don't need cfbd/cbbd codegen, they've already got libraries if we wanted them.'"* I worked from a stale copy of the backlog, saw "blocked on a free key", verified the key worked, and started — without re-reading the current entry first. That's on me.

The work is preserved on `feat/cfbd-codegen-stem` if it is ever wanted: 58 endpoints generated from the committed OpenAPI spec, 8/8 sampled wrappers verified live. Reopen rather than rebuild.

PR #455, opened 2026-09-03
feat(loaders): 12 loaders for published-but-unreadable NCAA tags

saiemgilani:
Twelve release tags have been published for years with no way to read them from Python. This adds the loaders — and corrects two stale claims about which gaps were real.

- **`ncaa_mbb_rapm` / `ncaa_wbb_rapm`** (16 seasons each, 2011–2026). The backlog said *"rapm has no Python loader"*; in fact `load_ncaa_{mbb,wbb}_rapm_within_team` already existed. It is the **base** tags that had none — so this is **2** loaders, not 4.
- **All ten `ncaa_mfb_*`** (pbp, pbp_cfbfastr, drives, schedule, rosters, teams, team_stats, player_stats, officials, linescore; 13 seasons each, 2013–2025) — the whole stats.ncaa.org men's football surface, unreadable from either language.

**Softball is deliberately not included: it is a phantom.** The backlog listed "softball (8) published-but-unloaded", but **zero softball tags exist** on the release (checked all 245 tag names — `ncaa_baseball_*` exists, softball does not). There was never anything to load.

**Schemas are derived from the published parquet, not hand-written** — one live read per tag, column names and dtypes taken from the file a caller actually gets, so the returns tables cannot drift from reality on day one.

**Verified by exercising every loader against the live release — 12/12 return data:**

| loader | rows | cols |
|---|---:|---:|
| `load_ncaa_mfb_pbp` | 349,065 | 51 |
| `load_ncaa_mfb_pbp_cfbfastr` | 305,268 | 105 |
| `load_ncaa_mfb_team_stats` | 202,344 | 10 |
| `load_ncaa_mfb_player_stats` | 59,781 | 28 |
| `load_ncaa_mfb_drives` | 39,717 | 17 |
| `load_ncaa_mfb_rosters` | 29,950 | 16 |
| `load_ncaa_mfb_linescore` | 13,530 | 11 |
| `load_ncaa_mbb_rapm` | 4,919 | 11 |
| `load_ncaa_wbb_rapm` | 4,557 | 11 |
| `load_ncaa_mfb_schedule` | 3,282 | 13 |
| `load_ncaa_mfb_officials` | 1,822 | 5 |
| `load_ncaa_mfb_teams` | 263 | 5 |

Codegen regenerated (loaders + parsed mirrors + reference docs); drift gate green.

## Summary by Sourcery

Make twelve previously inaccessible NCAA release datasets available through Python loaders.

New Features:
[...]

PR #459, opened 2026-09-03
docs(loaders): correct the season description on the remaining 59 loaders

saiemgilani:
Completes the fix started in #455, which corrected only the 12 loaders that PR added. The same wrong description was on **59 more**.

## The defect

`r_column_descriptions.yaml` defines `SEASON` — the NBA-stats column, which *genuinely* accepts a `'YYYY-YY'` string — and the R back-fill applies it **by column name** to every lowercase `season`. Across the generated reference docs that put *"4-digit year or 'YYYY-YY' string"* on:

- **59** `Int32`/`Int64` season columns (17 wbb, 30 wnba, 12 ncaa)
- **0** `String` season columns, where it would be correct

So it was wrong everywhere it landed. Passing `"2023-24"` to any of them raises `ValueError: invalid literal for int()`.

## The fix

Per-function entries in `manual_column_descriptions.yaml`, which wins over the R back-fill — the documented precedence, and where descriptions are supposed to live.

**One thing worth reviewing carefully:** 18 functions needed a new entry, but **41 already had an entry without a `season` key**. Those descriptions were *inserted into the existing block*, not appended as a new top-level key — appending a duplicate key would have made YAML load only the later occurrence and **silently drop every other column description in the original block**. Verified after the edit: 717 entries parse, no non-dict values (which is what a clobbered block would look like).

## Verification

Re-ran the measurement against the generated docs: **59 → 0** functions still carry the claim.

The measurement itself needed a correction mid-way — my first pass reported 71 because the detector was matching the *corrected* wording ("a 'YYYY-YY' string is **not accepted**"). Re-measured against the wrong claim specifically.

## Summary by Sourcery

Correct season parameter documentation across the remaining women’s basketball loaders to state that seasons require a four-digit starting year.

Bug Fixes:
- Correct season-column documentation across 59 women’s basketball loaders so integer season parameters are no longer described as accepting 'YYYY-YY' strings.

Enhancements:
- Add function-specific season descriptions to the manual column metadata so generated documentation accurately reflects loader behavior and preserves existing column descriptions.

Documentation:
- Regenerate WBB, WNBA, and NCAA WBB loader reference documentation with the corrected season format and validation guidance.

Tests:
- Verify the manual description metadata parses successfully and that no generated loader documentation retains the incorrect season claim.

## Summary by CodeRabbit

* **Documentation**
  * Clarified season field requirements across women’s basketball and WNBA loader references.
  * Documented that season values must be 4-digit starting-year integers.
  * Clarified that `YYYY-YY` string values are not accepted.


saiemgilani:
Closing unmerged rather than resolving the rebase conflict, for a specific reason:

This branch's history carries a divergent duplicate of #455's already-merged commit
(same content, different hash), so a rebase onto current main conflicts on that
already-landed change, not on anything genuinely new here.

Compared this PR's 59-schema fix against #460 (merged separately, same underlying
bug) directly via a YAML diff: 47 of the 59 schemas overlap exactly, byte-identical
wording. Of the 12 unique to this PR, all are `load_ncaa_wbb_*` loaders — extracted
and landed cleanly in #461. The 6 schemas unique to #460 (ESPN direct wrappers) were
not in this PR's scope, so nothing here is lost.

Thanks for finding and fixing this — the underlying bug (a mined R-package
description meant for one function applied by name-only matching regardless of the
consuming schema's actual contract) was real, and this PR's diagnosis of "the
remaining 59" independently matched what a separate investigation also found.

PR #466, opened 2026-09-07
feat(cfb): score cp/cpoe with an air-yards booster where air yards exist

saiemgilani:
The consumer half of [cfbfastR-cfb-data#73](https://github.com/sportsdataverse/cfbfastR-cfb-data/pull/73), which trains and publishes the model. This bundles it and puts it in front of game-on-paper, which reads CPOE from `cfb_pbp.__process_cpoe`.

## Why

The bundled 8-feature game-state CP model is close to a coin flip at the thing it exists to do. On the plays that carry air yards:

| arm | log-loss | AUC |
|---|---:|---:|
| intercept-only (base rate 0.627) | 0.6603 | 0.500 |
| **game-state only (what shipped)** | **0.6539** | **0.567** |
| **game-state + air yards** | **0.5398** | **0.764** |

_5-fold CV grouped by `game_id`, 27,804 plays, 2025–26._

It beats an intercept by **0.0064 of log-loss**. That was the entire skill of CFB `cp`, and therefore of CFB `cpoe`.

It shows in the predictions themselves — scoring real 2025–26 plays:

| arm | n | mean cp | actual | **sd(cp)** |
|---|---:|---:|---:|---:|
| air_yards | 25,057 | 0.6246 | 0.6245 | **0.2249** |
| game_state | 30,208 | 0.5811 | 0.5472 | **0.0646** |

The old model emits approximately the base rate for every play regardless of down, distance or field position.

## What changed

- Bundles `cfb_cp_model_air_yards.ubj` (11 features = `CP_FEATURES` + `air_yards`, `pass_is_middle`, `qb_hurry`). Ships via the existing `cfb/models/*` package-data glob — no `pyproject.toml` change.
- `__process_cpoe` now routes each play to whichever model can see it and records that in a new **`cp_model`** column.
- `pass_is_middle` is left **null** when `pass_direction` is null rather than filled 0 — "not a middle throw" is a different claim from "we don't know", and xgboost reads NaN as missing rather than as a zero.

## Both models stay

ESPN only emits the catch/target spots `air_yards` is derived from in 2025+ — 38.9% of 2025, 90.2% of 2026 to date, ~none before. The game-state booster remains the only thing that can score the overwhelming majority of the historical corpus, and game-on-paper renders historical games. The air-yards arm is never extrapolated onto plays whose throw distance is unknown.

⚠️ **`cp` from the two arms is not on a comparable scale.** Anything aggregating `cpoe` must group by `cp_model` or restrict to one — coverage climbs steeply season over season, so a blended average would read that composition change as a trend. The docstring says so.

## Verification

- Live 2026 game through the full pipeline: 57 pass plays, **40** routed to the air-yards arm and **17** to game-state, all 108 non-pass rows null.
- **814 CFB tests green**; codegen drift gate clean (nothing regenerated).
- 4 new tests, including one that asserts throw depth actually *moves* `cp` (a 1-yard vs a 45-yard throw in identical game state must differ by >0.10) — a wired-but-inert feature is exactly the failure mode the game-state model had.

## Note on the second commit

`chore(deps): lock the nflpro extra` is **unrelated** to this work. `pyproject.toml` on main declares an `nflpro` extra that `uv.lock` never picked up, so the mypy pre-push hook re-locks it on every push and then fails for having modified a file. Kept as its own commit rather than folded into the CPOE change; happy to split it out if preferred.

## Follow-up

Once cfbfastR-cfb-data republishes `espn_cfb_model_pbp` with the new `cp_model` column, `loader_schemas.yaml` + the returns-table description for `load_cfb_model_pbp` will need the entry. Not actionable until that data lands.

## Summary by Sourcery

Improve college football completion-probability scoring by routing eligible plays through an air-yards model while preserving a consistent game-state series for aggregation.

New Features:
- Add an air-yards-enhanced completion-probability model for plays with known throw depth and identify the selected model per play.

Bug Fixes:
- Preserve unknown pass direction as missing data instead of treating it as a non-middle throw.

Enhancements:
- Retain the original game-state completion probability as a consistent aggregation series and use it for box-score expected completions and CPOE calculations.

Documentation:
- Document the different scales and intended aggregation use of hybrid completion probability versus game-state completion probability.

Tests:
- Add coverage for model bundling, per-play routing, historical fallback behavior, throw-depth sensitivity, and aggregation-scale consistency.

Chores:
- Lock the previously unresolved nflpro dependency extra in the project lockfile.

## Summary by CodeRabbit

* **New Features**
  * Improved college football completion-probability estimates by accounting for throw depth and additional play context.
  * Added `cp_model` and `cp_game_state` fields to distinguish prediction inputs and preserve the original game-state estimate.
  * Updated passer expected-completion totals to use consistent game-state probabilities across pass plays.
  * Retained game-state estimates when throw-depth information is unavailable.

* **Tests**
  * Added coverage for model selection, fallback behavior, feature availability, depth-sensitive probabilities, and aggregation consistency.


saiemgilani:
Two findings; one fixed, one declined.

**`create_box_score` blending the two CP scales — fixed, and a genuine miss.** The docstring I wrote warns consumers not to aggregate `cpoe` across models, and then `create_box_score` in the same file does exactly that: `xComp = cp.sum()` into passer `CPOE`. A passer with both air-yards and fallback plays would have had two different quantities summed, and because coverage varies by game, passers would also have stopped being comparable to each other.

Fixed by separating the two jobs rather than by grouping:

- **`cp`** — the hybrid, best available per play. What a play view should show.
- **`cp_game_state`** (new) — the game-state prediction on *every* pass play, one consistent scale. What anything summed or averaged should use.
- **`cp_model`** — provenance.

`create_box_score` now sums `cp_game_state`, which also means `xComp` / `CPOE` keep exactly the meaning they had before this PR — existing consumers see no change, while per-play `cp` gets better. Grouping the box score by `cp_model` was the other option but it would have changed the output shape (multiple rows per passer) for every consumer.

Added a test asserting `cp_game_state` is non-null on every pass play, agrees with `cp` on fallback rows, and *diverges* from it on air-yards rows — that last assertion is there so a wired-but-inert air-yards arm fails the test rather than passing quietly.

**E712 on lines 7648–7649 — declining.** This is deliberate project convention, not an oversight: `pl.col("c") == True` is the documented style for polars boolean masks in this repo, `E712` is suppressed in `pyproject.toml` (line 278), and CLAUDE.md records the reasoning (it is clearer when the column is itself an expression, and avoids surprises around null handling). `ruff check` passes clean on the file. The surrounding code in `__process_cpoe` already uses `== True` throughout, so changing only the new lines would make them inconsistent with their immediate neighbours.

PR #520, opened 2026-09-17
fix(nfl): pin fantasy-football id columns as strings instead of inferring them

saiemgilani:
## What

`load_nfl_ff_playerids` and `load_nfl_ff_rankings` read the DynastyProcess CSVs with **inferred** dtypes. Each id column's type therefore depended on whatever rows upstream currently had in polars' 100-row inference window:

- `fantasypros_id`, `pff_id` and `nfl_id` are documented as character but now come back as Int64;
- `yahoo_id`, `fleaflicker_id`, `rotoworld_id` and `swish_id` only stay String because they are null in the first 100 rows;
- Int64 drops zero-padding: live `nfl_id` `"038666"` becomes `38666`, and 225 `mfl_id` values like `"0156"` lose their padding.

This PR pins the id columns at read time with `schema_overrides`, so they're parsed from the CSV text and never go through a float:

- **playerids:** all 20 `*_id` columns are `Utf8`.
- **rankings, `kind="draft"`:** `id` (the FantasyPros id; 92% of live values appear in playerids `fantasypros_id`), `sportsdata_id`, `yahoo_id` and `cbs_id` are `Utf8`.
- **rankings, `kind="week"`:** `fantasypros_id` and `player_opponent_id` are `Utf8`.

## Why all Utf8 and not the documented types

The "integer" types in the autodoc schema were just what polars inferred when the schema was captured. The evidence for strings:

- upstream's own `db_playerids.rds` stores all 20 ids as character (checked with `readRDS`);
- the `db_fpecr.parquet` "all" asset already types `id` as String;
- `nfl_roster_builder._SCHEMA` and `nfl_players._SCHEMA` pin `espn_id`, `sleeper_id`, `rotowire_id` and `fantasy_data_id` as `Utf8`, so a roster -> playerids join on those keys raised `SchemaError` whenever playerids inferred Int64;
- an Int64 pin raises on the first non-numeric value, while Utf8 loses nothing.

## Caller-visible change

Id columns that currently infer as Int64 will come back as String, in both polars and `return_as_pandas=True`. Every change is listed in `CHANGELOG.md` under Unreleased, with a `clear_cache()` hint for filesystem-cache users.

## Evidence

- **Fixtures:** verbatim slices of all three live CSVs; provenance is in the fixture README.
- **Tests:** they failed on the pre-fix loader (7 id columns typed Int64, then 5/6, then 4/11 as each commit landed) and all pass now. The cross-loader test imports the real `_SCHEMA` dicts and doesn't restate them.
- **Live check:** against the full files, every pinned read parses and only the intended columns change dtype. The rankings -> playerids joins match 4,537 draft rows and 756 weekly rows.
- **Docs:** `generate.py` + `--check` are clean, with no generated-doc changes. The returns tables render these id types from the autodoc schema, which `chore/autodoc-refresh` updates to `character`. Rebased onto #507, with both CHANGELOG entries kept.

## Summary by Sourcery

Preserve DynastyProcess fantasy-football identifiers as strings across supported loaders and outputs.

Bug Fixes:
- Pin DynastyProcess fantasy-football player and ranking identifiers to string types so upstream row ordering cannot change their dtypes or strip zero-padding.
- Keep fantasy-football identifiers consistent across loaders, preserving joins with roster/player data and between rankings and player IDs.

Documentation:
- Document the identifier dtype change, affected loaders, cache invalidation guidance, and required casting for integer-based callers in the changelog.

Tests:
- Add offline regression coverage using real DynastyProcess CSV slices for identifier dtypes, zero-padding, cross-loader joins, schema compatibility, and pandas output.

Chores:
- Add captured DynastyProcess CSV fixtures and provenance documentation for dtype regression testing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fantasy football player and ranking loaders now consistently treat identifier columns as text, preserving leading zeros, null values, and alphanumeric IDs.
  * Pandas output maintains string-typed identifier columns for consistent downstream use.
  * Existing cached data may retain previous types until expiration; clear the cache after upgrading to apply the updated behavior.

* **Documentation**
  * Added changelog details covering affected loaders, identifier formats, and cache considerations.


saiemgilani:
Addressed CodeRabbit's outside-diff note in 8c739bbd5: the `tests/fixtures/nfl_loaders/README.md` intro now describes both the nflverse parquet slices and the DynastyProcess CSV slices. Sourcery skipped this PR because the weekly review budget is used up.

PR #523, opened 2026-09-17
fix(mlb): return Statcast search MLBAM id columns as Int64, not Float64

saiemgilani:
## What

`mlb_statcast_search`, `_minors` and `_wbc` returned the runner ids `on_1b` / `on_2b` / `on_3b` as **Float64**, e.g. `660271.0`.

`_csv_to_frame` reads the Savant CSV with `pd.read_csv`, which turns any integer column containing a blank into float64. `on_1b` is blank whenever first base is empty. Float ids break joins against `batter` / `pitcher` (Int64), and they stringify as `"660271.0"`.

`_csv_to_frame` now pins the 14 MLBAM id columns to nullable **Int64**: `batter`, `pitcher`, `on_1b`..`on_3b`, `fielder_2`..`fielder_9` and `game_pk`. That turns `660271.0` into `660271` and a blank into null.
- **Scope of the pin:** it only touches id columns that were read as floats. Leaderboard CSVs that share the helper keep their dtypes, as does every non-id column.
- **Non-integral values:** a column holding one is left as read, with a single `UserWarning` per call attributed to the caller's line, so nothing is truncated silently.
- **pandas output:** `return_as_pandas=True` gives nullable `Int64` for all 14 ids, from both the parser and the search function.

## Caller-visible change

`on_1b` / `on_2b` / `on_3b` change from Float64 to Int64. In pandas, `batter`, `fielder_*` and `game_pk` from `parse_mlb_statcast_search` change from numpy `int64` to nullable `Int64`. The Returns docs change from `character` to `integer`. See the CHANGELOG entry.

## Evidence

- **Fixture:** `search_2024-06-15_head.csv` is the header plus the first 46 rows of a real `/statcast_search/csv` response (provenance in the fixture README).
- **Exact values:** the tests assert the bases-loaded pitch (game 745329, at-bat 44, pitch 2) gives exactly `(656305, 671218, 596103)` as Python `int`, in both polars and pandas.
- **Multi-chunk:** a two-chunk test stitches an all-blank `on_3b` chunk onto the real fixture. On origin/main this came back Float64.
- **Fail-first:** the new tests failed on origin/main's parser and search module, and all pass now: 31 statcast parser, search and codegen tests, with the drift-gate test deselected.
- **Live check:** a review verified against five live Savant responses (2024-06-15 batter and pitcher, 2015-06-15, minors, WBC). All share the 119-column header, and over 15,419 rows all 14 ids equal their float originals with nulls preserved. The four downstream analytics modules already cast to Int64, and their tests pass.
- **Gates:** `ruff` and `mypy` are clean. `generate.py` + `--check` are clean, and the regenerated MLB reference changes 9 runner-id rows to `integer`. Rebased onto #507.

## Merge order

#521 has merged. main was merged into this branch: the three `mlb_statcast_search*` autodoc schemas and the MLB reference page take #521's versions, which already type all 14 ids as `integer`.

## Found in review, not changed here

- **Minors and WBC routes:** with no filters, `mlb_statcast_search_minors` returned MLB regular-season rows, and `_wbc` for 2023-03-11 returned spring training.
- **`game_pk` dtype mismatch:** `parse_mlb_statcast_gamefeed` returns `game_pk` as String while search returns Int64, so a join between them fails.
- **Header-only CSV:** returns a (0, 0) frame instead of the documented columns.

## Summary by Sourcery

Fix MLB Statcast search identifier handling so MLBAM IDs use nullable integer types consistently without silently altering invalid non-integral values.

Bug Fixes:
- Return MLB Statcast search MLBAM identifier columns as nullable Int64 values, preserving nulls for blank CSV fields and preventing float-formatted IDs from breaking joins or stringification.
- Leave non-integral identifier values unchanged and emit a single caller-attributed warning instead of silently truncating them.

Enhancements:
- Apply consistent identifier dtypes across MLB, minors, and WBC Statcast search outputs in both Polars and pandas.
- Preserve identifier dtypes when concatenating date chunks and converting search results to pandas.

Documentation:
- Document the Statcast identifier dtype correction and updated integer return types in the changelogs.

Tests:
- Add fixture-based parser and search coverage for exact integer IDs, null runner IDs, multi-chunk concatenation, pandas output, and non-integral ID warnings.

## Summary by CodeRabbit

- **Bug Fixes**
  - MLB Statcast search results now preserve MLBAM identifier columns as nullable `Int64` values in Polars and pandas outputs.
  - Blank runner identifiers remain null instead of causing identifier columns to become floating-point values.
  - Identifier values remain exact, improving compatibility with joins and downstream processing.
  - Non-integral identifier values are preserved without truncation and generate a single warning per request.

- **Documentation**
  - Updated return-type documentation and changelog details for MLB Statcast identifier columns.


saiemgilani:
CodeRabbit nitpick (move the Statcast test modules to tests/mlb/): not done here. `tests/test_mlb_statcast_parsers.py` and `tests/test_mlb_statcast_extra.py` already existed at that path, and this PR only adds tests to them. Moving them is a layout change unrelated to the dtype fix, so it's left for a separate cleanup.

PR #532, opened 2026-09-18
fix(cfb): Stage-1 tail — shared Last,First name shape, derived stepped-out punt returns, text_dupe description

saiemgilani:
Stage-1 tail for CFB: the three follow-ups the S1-CFB slice parked (LEDGER C26, C42,
and the stale `text_dupe` description). One commit each.

## C26 — the vendor special-teams name regexes move into the shared football grammar

`cfb_pbp.py` still carried `_VENDOR_FG_KICKER_RE`, `_VENDOR_KICKOFF_RETURNER_RE` and
`_VENDOR_PUNT_RETURNER_RE`. They are replaced by **one shared name expression** in
`football/espn_text.py` -- `CLAUSE_NAME` -- plus the two anchors built from it:
`CLAUSE_RETURNER_RE` (one pattern for punts and kickoffs, since each call site already
knows which kick the row is) and `CLAUSE_FG_KICKER_RE`. The CFB processor now holds no
name regex of its own for these clauses; it reads both through `jersey_name()`.

**The ledger's premise did not survive the replay.** C26 assumed the locals only added
the `"Last,First"` shape, so the plan was a precise surname-first expression composed
with `ABBREVIATED_NAME`. That was built and measured first, and it is a regression:
replayed over every local ESPN summary (3,227,541 plays / 20,718 games, 2004-2026) it
**loses 57 names and rewrites 20** the locals get today --

* 20 FG kickers in 2005 whose text spells the name out ("Bryan Hahnfeldt field goal
  attempt from 34 GOOD" -- no jersey, no initials, no comma), plus 37 punt returners of
  the same shape in 2005-2014 ("Joe Radigan punt 48 yards to the UConn17, Brandon McLean
  return 32 yards to the UConn49");
* 8 kickoff returners changed, 3 of them to the **wrong player** -- dropping the jersey
  requirement let the abbreviated pattern match a later return clause in the same row
  ("Cameron Echols-Luper return for 8 yds ... B.J. Catalon return for 30 yds" read
  B.J. Catalon).

The loose capture is what carries those names, and the kick-clause anchor is what makes
it safe -- so the consolidation keeps that design and moves it, rather than narrowing it.

**Proof of no loss** (the 2025-26 store and every other season): base vs this branch over
the same 3,227,541 plays --

| column | non-null base | non-null new | lost | changed | gained |
|---|---|---|---|---|---|
| `punt_return_player_name` | 98,464 | 98,464 | 0 | 0 | 0 |
| `kickoff_return_player_name` | 101,247 | 101,247 | 0 | 0 | 0 |
| `fg_kicker_player_name` | 52,495 | 52,495 | 0 | 0 | 0 |

End-to-end through `run_cleaning_pipeline` on the 4 games the narrow variant broke: 656
plays, 0 cells differing on all three name columns.

7 new verbatim cases in `tests/football/test_espn_text.py` (each quoted from a stored raw
summary, id in the case name). Mutation: requiring the jersey in `CLAUSE_NAME` turns 5 of
them red.

## C42 — the punt return whose text ends at "out of bounds" with no yardage

Since C38b stopped reading a post-return "out of bounds" as the KICK going out, a punt
whose only stated outcome is the returner stepping out has a **null** return rather than
the old 0 -- and #508's derivation skipped those rows because `_PUNT_OUTCOME_RE` contains
`return`. It now admits exactly that case (text matches `_RETURNER_STEPPED_OUT_RE`), with
#508's own formula and guards and the same `yds_punt_return_source="derived"` stamp.

1,329-game 2004-2026 sample (236,924 plays), base vs new:

* 41 punt rows in the sample are a stepped-out return with a null yardage (the ledger's
  42, all 2008-2013 -- the era of that template); **25 of them derive**, the other 16 are
  held back by #508's own guards (the next snap does not start at the end spot, the result
  is exactly 5 or 15, or it is not positive).
* `yds_punt_return`: **25 null -> derived**, 0 parsed values changed, 0 emptied; all 25
  stamped `"derived"`.
* `yds_punted`, `yds_kickoff`: 0 rows differing.

3 new tests (real 2008 row + a stated-yardage control + the other stated outcomes that
must stay undrivable). Mutation: dropping the new branch turns the first red.

## `text_dupe` description

`gen_cfb_pbp_descriptions.py:145` still said "True when the play description duplicates
the previous row's text". After C39 the dupe rows are filtered **before** the frame is
returned, so the column is always `False`. Corrected in the generator and in
`manual_column_descriptions.yaml` (the source of truth for the returns table) to say so.

## Tests

`tests/cfb` + `tests/football` offline, before and after: base tree 1,272 passed / 0 failed (5 errors are a fixture-copy artifact: the baseline copy has no gitignored `dev/cfb_projection/`), branch **1,282 passed, 56 skipped, 2 xfailed, 0 failed** -- the 10 new tests are the delta. CI `pytest (offline, parallel)` green.

`ruff format --check` / `ruff check` / `mypy sportsdataverse/football/espn_text.py` clean.

**Codegen**: `generate.py --docs` was run and rewrote exactly one line of one file --
`docs/docs/cfb/reference/loaders.md` (the `text_dupe` row) -- which is committed here, so
the drift gate has nothing left to flag from this branch. No public function, docstring
signature or `__all__` changed; `espn_text` gained three module constants (`CLAUSE_NAME`,
`CLAUSE_RETURNER_RE`, `CLAUSE_FG_KICKER_RE`) and lost none.

Rebased onto `ac5b33067` (#533) so `ruff + mypy` is green -- the earlier red was that PR's
pre-existing `tests/codegen/test_generate.py` format defect on the base, not this branch.

Evidence, replay scripts and mutation patches: `background-research/2026-09-17-football-sources-program/tail-cfb/`.


saiemgilani:
Heads-up, not from this branch: `ruff + mypy` fails on `ruff format --check tests/codegen/test_generate.py`, which is unformatted on `integration/football-sources` itself (last touched by #513's rebase, 5442534c5). This PR touches none of `tools/codegen/generate.py`'s tests — `ruff format --check` and `ruff check` are clean on all seven files it does touch — so the gate needs that one file reformatted on the base branch (or by whoever owns the codegen slice) rather than here.

PR #518, opened 2026-09-17
fix(cfb): S1-NCAA round 3 (NC1-NC6, NC8)

saiemgilani:
Round 3 of the S1-NCAA slice, stacked on #514 (`fix/s1-ncaa-r2`; #502 was merged as `bd3a4cc88`). Ledger items **NC1-NC6** fixed, **NC8** decided and documented, **NC7** report-only, plus two more bugs the round surfaced. Same rules: real fixtures, mutation-checked tests, nothing published.

## NC5 - replay reprints (`cf0136dc9`)
The review note ("The previous play is under automatic review ... PLAY OVERTURNED. (Original Play: ...)") is cut once before decomposition, so every flag, yardage, end spot, tackle and penalty comes from the ruling; `play_text` keeps the whole string. Round 2 had only patched `is_touchdown`.

## NC3 - both name orders (`dcfd12a8c`)
`_NAME` accepts `Last[ Suffix],First` (a bare initial allowed: `McKENZIE, D`), 2019-era `First Last` (`Joe Burrow`, `C. Ed.-Helaire`, `K. Duncan Jr.`, a lone `ALEXANDER-STEVE`) and the 2025 `#95 K.Kimble` jersey form, case-sensitively inside the case-insensitive play regexes so narrative words can never pass for a name. Route words between passer and "pass" (`Liam Welch middle pass complete`), the 2019 formation codes (`SH,` `SHOT,` `SG,` `SGUN,` `NHSG,` `NH,` `PSTL,`) and the `Clock MM:SS,` prefix are handled; the swallowed sentence period (`DRAYTON, Matt.`) is dropped. `to_cfbfastr` still emits `First Last`: comma names are turned around, the others pass through, the jersey prefix is removed.

## NC4 - tackler separators (`4d7c75d6b`)
`Last, First` groups split on `;`, on the mangled `:` (`MORGAN, D.J.3aPAUL, Keyshawn`) and at the `Last, ` that starts the next name (a cut inside a CamelCase surname is glued back); every piece must be a name or the group is dropped. Comma-less groups: the glue between two `First Last` names (`Kristian FultonJaCoby Stevens`) is indistinguishable from a CamelCase first name (`TreShun`), so only a single `First Last[ Jr.]` is trusted and the rest are **left null** (deterministic, not garbage).

## NC2 - TD and PAT in one row (`968f491b9`)
[...]

PR #502, opened 2026-09-17
fix(cfb): stop kickoff rows mirroring NCAA yards_to_goal

saiemgilani:
## Problem (ledger item C1)

`to_cfbfastr` emitted `yards_to_goal` / `yards_to_goal_end` mirrored (100 - true) for most NCAA games. In the published `ncaa_mfb_pbp_cfbfastr_2025` asset, 4,177 of 5,724 rushing TDs have `yards_to_goal == 100 - yards_gained`, across 1,078 of 1,628 games. QA and tests only checked final scores, so nothing caught it.

## Root cause

`_own_side` (`cfb_ncaa_cfbfastr.py:228`) works out which yard-line side code belongs to each team. It votes on the first row of every drive, on the theory that drives start in the offense's own territory. On stats.ncaa.org that first row is usually the **kickoff**. The kickoff sits in the **receiving** team's drive but is spotted on the **kicking** team's side (`HOU35` in an Oregon St. drive). Whenever drives that start with a kickoff outnumber the rest, the vote flips, and every `yards_to_goal` in the game is mirrored.

## Fix

Only scrimmage plays vote now (`rush`, `pass`, `sack`, `kneel`, `punt`, `field_goal`). Drives with no scrimmage play (kickoff, PAT or penalty only) cast no vote. The four heuristics in the module docstring are unchanged: the FCS defense-labelled-drive vote, the checkpoint-slot vote, linescore score snapping and OT synthesis. No public signature, docstring or `__all__` changed, so no codegen.

## Tests

The new field-position tests run on 7 real games. Five of them were mirrored before the fix: 5336803, 5361446, 5362431, 6386335 and 6386574. 5362535 and 6386512 were not.
- Rushing and receiving TDs (no penalty, no turnover) have `yards_to_goal == yards_gained` and `yards_to_goal_end == 0`.
- After a kickoff touchback, the next snap has `yards_to_goal == 75`.
- `yards_to_goal` is filled in for every snap that has a yard line, and stays within [0, 100].
- 6386512 (Houston @ Oregon St., 1OT) runs through the **full bundle**: pbp, drive titles, drives, linescore and scoring summary, called the same way the `-data` build calls it. It checks the 27-24 final and that the OT rows are synthesized. New fixture: `mfb_play_by_play_6386512.html`, taken byte-for-byte from the producer bundle. The box score and drives tabs for this game were already committed.

[...]

PR #536, opened 2026-09-18
feat(nfl): shield_pbp handles in-progress games (Stage 2 Phase 2)

saiemgilani:
Stage 2 Phase 2 of `background-research/2026-09-16-nfl-shield-live/PLAN.md` §4: make the
graduated Shield parser (`sportsdataverse/nfl/shield_pbp/`, #528) correct on a game that is
still being played. Evidence is the 2026 wk2 DET@BUF TNF capture — 223 `gamedetails_v2`
snapshots 45 s apart, pregame through final.

## The six live bugs

| # | Bug (as found on a real in-progress payload) | Fix |
|---|---|---|
| 1 | The open drive (`endedPlaySequenceNumber: null`) was dropped from the play→drive assignment, so every play of the drive being played had `posteam` null — and with it `defteam`, `yardline_100`, `score_differential`, `home`, and garbage EP | `parse._drive_ranges` gives an open drive a synthetic end: the next drive's start, or `+inf` when it is the last drive |
| 2 | `fixed_drive` never advanced past the last closed drive | falls out of #1 — the cascade keys on a possession change, which needs `posteam` |
| 3 | `home_score` / `away_score` / `result` stamped the **current** score as the final one | `add_labels` holds all three null until `summary.phase` is `FINAL` / `FINAL_OVERTIME` |
| 4 | No row for the next snap, although `summary` carries down / distance / yardLine / clock / timeouts | `live.current_situation_row` appends one, flagged `is_play = 0`, `shield_play_type = "CURRENT_SITUATION"`; scores and timeouts come from `summary`, not the play-derived running totals |
| 5 | The provisional tail was unmarked | `live.add_live_columns` adds `provisional` for the trailing run of plays the feed has not closed (`playEndTime` null) |
| 6 | `roof` / `spread_line` / `total_line` only ever came from a nightly schedules join | `live.resolve_context`: caller-injected → nflverse schedule row → the repo-wide 2.5 / 55.5 processor default |

## Entry point

```python
shield_nfl_pbp(game_detail | shield_game_id, *, enrich=True, context=None)
```
→ an `nfl_model_pbp`-shaped frame for any phase. `game_detail=` keeps pollers and tests off
the network; `shield_game_id=` fetches through
`nfl_game_details_v2(include_drive_chart=True, return_parsed=False)`. New columns
`live_phase` / `is_play` / `provisional` are added by this entry point only — `build_pbp`,
and therefore the published `nfl_model_pbp` schema, is untouched.

## Gates

Full detail in `background-research/2026-09-17-football-sources-program/s2-shield-live/GATE.md`.

| gate | result |
|---|---|
| **Finals frame equality** — 3 real nfl-raw 2025 wk1 finals, `build_pbp` before vs after | **exact**, (154/162/180, 233) each, no new / dropped / differing column |
| **Prefix invariant** — 223 snapshots, every play whose raw payload object is unchanged vs the final payload | **0 base-column mismatches** over **19,820** row comparisons |
| **EP/WP in range** — same 223 snapshots, enriched | **20,477** scored rows, **0** outside `ep` ±7 / `wp`, `vegas_wp` in [0, 1] |
| **No null `posteam` on live plays** | **0** (was 8 of 90 rows on the Q2 snapshot) |
| **Mutation sweep** — 5 mutations, one per fix | **5/5 caught** |
| Tests | `tests/nfl/shield_pbp/` 134 → **154**; ruff clean; `generate.py` + `--check` clean |

62 of the 19,820 enriched rows (0.31%) differ in an *enrichment* column only — `wpa` / `epa` are
lead-diff (they need the next play), `total_*` are running sums, `ep` depends on
`receive_2h_ko`, which needs a non-null `defteam` to exist anywhere in the frame. Histogram in
`out/gate_enrich2.log`.

One unrelated commit rides along: `docs/docs/cfb/reference/additional.md` was stale on `main`
(PR #534 changed `yahoo_cfb_boxscore`'s default without regenerating), which made the codegen
gate red independently of this branch. Drop that commit if another slice regenerates it first.

## Summary by Sourcery

Make NFL Shield play-by-play parsing accurate and poll-safe for in-progress games while preserving final-game behavior.

New Features:
- Add `shield_nfl_pbp` as a live-capable NFL Shield parsing entry point for any game phase, including offline payload input, optional enrichment, and injectable game context.
- Expose live phase, play-row, provisional-state, and current-situation data for in-progress games.

Bug Fixes:
- Preserve possession and drive assignments for open in-progress drives.
- Defer game outcome fields until Shield reports a final phase instead of using live scores.
- Add the next-snap situation row and mark unsettled trailing plays as provisional.
- Provide robust context resolution through caller values, schedule data, and documented defaults.

Enhancements:
- Keep the published `build_pbp` and `nfl_model_pbp` schema unchanged while layering live-specific fields at the new entry point.
- Handle empty and pregame payloads consistently, including stable live-column schemas and safe context fallbacks.

Documentation:
- Regenerate the CFB Yahoo boxscore reference documentation to reflect its parsed-by-default behavior.

Tests:
- Add real Shield fixtures and comprehensive live-game coverage for open drives, final-state labels, situation rows, provisional plays, context resolution, empty payloads, and cross-snapshot prefix stability.

Chores:
- Correct stale generated CFB reference documentation included alongside the NFL changes.

## Summary by CodeRabbit

* **New Features**
  * Added live NFL play-by-play support, including game-phase detection, provisional play indicators, current-game situation rows, and live context such as scores, possession, and timing.
  * Improved in-progress drive and possession data for live games.
  * Added support for building live play-by-play data from game details or a game ID, including optional advanced metric enrichment.

* **Documentation**
  * Updated college football box score documentation to reflect parsed data as the default return format.


saiemgilani (commented):
## Review — the live layer is right; four failure paths the gate could not see, now fixed in-branch

Reviewed at `1c311c9fa`; fixes pushed as `9c6897f5c` + `65101fc08`, everything run under bubblewrap with `--unshare-net`. **The five live fixes are correct and mutation-verified, and a final's frame does not move** — I reproduced GATE 1 independently on `2025_01_DAL_PHI`: `a.equals(b)` is `True`, 154×233, same columns, order and dtypes as `main`. `build_nfl_season(source="shield")` cannot see the `is_play=0` row either: it routes through `shield_pbp.build.build_season` → `build_pbp` (`nfl_build.py:292`), never `shield_nfl_pbp`. And `shield_nfl_pbp` rebuilds from the payload on every call, so the ~11-plays-back revisions are handled by construction — it never appends.

I re-ran two of GATE 4's mutations myself and both reproduce (m1 → 3 red, m2 → 1 red).

What the 223-snapshot sweep could not see is that `gate_prefix.py:72` does `if df.height == 0: continue` and always passes a full `context=`. Everything below lives in that blind spot. Fixed on the branch in two commits, each mutation-verified.

### 1. A not-yet-played game returned a 0-**column** frame · major · fixed

The state **every** scheduled game is in, so a poller started before kickoff hits it first — all 256 unplayed 2026 games in `nfl-raw` are exactly this (`phase: null`, 0 plays):

```
shield_nfl_pbp(game_detail={"driveChart": {"plays": [], "drives": []}}, enrich=False, context=CTX)
→ height 0   live cols present: []   ncols 0
```

`build_pbp` returns `pl.DataFrame()` and the early `return df` handed it straight back, so the three columns the returns table promises were missing and `df.filter(pl.col("is_play") == 1)` was a `ColumnNotFoundError` rather than an empty result. CLAUDE.md: empty frames carry the documented schema.

Fixed by declaring them through the **schema**, not `with_columns`/`select` — a literal added to a 0×0 frame yields a **one-row** frame in polars 1.42, which would invent a play. My first attempt did exactly that and the new test caught it.

Same commit guards `{"data": null}`, which is what the fetch returns for a uuid Shield does not know: `game` became `None` and crashed with an `AttributeError` three calls deeper. `mypy` flags both — but `shield_pbp` is outside the `[tool.mypy] files` ratchet, so CI never saw it.

### 2. The context resolver fabricated the market inputs in silence · major · fixed

```
resolve_context(game, None, game_id=None) → ('outdoors', 2.5, 55.5)    warnings: []
```

No warning, no provenance, and `_game_id_of` swallows every exception and returns `None` — so a payload whose game_id will not parse takes the same path without even attempting the schedule. `spread_line` / `total_line` are the only inputs to `vegas_wp`, and nothing in the frame distinguishes a defaulted line from a real one, so the live WP surface shows a "Vegas-informed" number off a made-up line. Measured on the situation row: the real context gives `vegas_wp` 0.159; the defaults move it.

Fixed with one `RuntimeWarning` naming every field that fell back. **A `context_source` column would be stronger** and matches the provenance columns the CFB slices of this program already ship — your call, since it is a schema addition.

### 3. The headline invariant had no committed test · major · fixed

The strongest evidence in the GATE — 0 base-column mismatches over 19,820 rows / 223 snapshots — is produced by a script in `background-research/`, which CI never runs and nobody without the capture directory can reproduce. The two committed fixtures were one INGAME snapshot and a 1-play PREGAME shell of the same game, so no committed test could express "a later poll must not move a row whose raw payload object has not changed". A regression in the open-drive synthetic end, or in any frame-global derivation, would have shipped green — the same shape as [#502/#518's `_own_side` nondeterminism](https://github.com/sportsdataverse/sportsdataverse-py/pull/518), which only surfaced because someone diffed two runs.

Fixed by committing a second snapshot of the same game (152/223, Q4 15:00, 143 plays, 99 of the Q2 fixture's 103 byte-identical, 35 KB) plus `test_prefix_invariant_across_two_snapshots`, mirroring `gate_prefix.py`'s scoping and exclusion set exactly. Mutation-verified: numbering `fixed_drive` from the end of the frame turns it red.

### 4. The claim justifying the phase gate is false — and one archived game's published output moves · minor · comment fixed, **data call open**

`labels.py` says "every nfl/raw payload 1999-2026 carries phase FINAL or FINAL_OVERTIME". Measured over the whole library:

```
non-PRE payloads 7,549 → FINAL 6,846 · FINAL_OVERTIME 445 · None 257 · PREGAME 1
of the 258 non-FINAL: 257 build 0 play rows; 1 builds 2 rows → 2022_17_BUF_CIN (PREGAME, 7-3)
```

So `build_nfl_season(seasons=[2022], source="shield")` publishes that game with `home_score`/`away_score`/`result` going 7 / 3 / 4 → null / null / null. GATE 1's three games are all FINAL, so the equality gate is structurally blind to this class.

**This is the right direction, not a regression**: nflverse's own 2022 schedule has no row for `2022_17_BUF_CIN` at all, so a null outcome is closer to parity than a stamped 7-3, and a null `result` drops the row from model training rather than training on a wrong margin. I corrected the comment to the real inventory. **Your call:** should `build_season` skip the abandoned game entirely? nflverse has neither its schedule row nor its 2 plays.

### 5. `test_schedule_lookup_failure_degrades_with_a_warning` reached the network · minor · fixed

It called `load_nfl_schedule([1899])` for real and passed off whichever error the environment happened to raise — offline a DNS failure, online something else. `tests.yml:86` states the rule this breaks: *"Verified offline under bubblewrap with --unshare-net: 0 network-reaching tests as of 2026-09-18. If you add a test that fetches, gate it in conftest.py or mock the fetch."* Now monkeypatched, like its sibling two tests up.

### 6. `provisional` could be null, or silently all-zero · minor · fixed

`is_in` propagates nulls in polars 1.x, so a row with a null `play_id` got `provisional = null`, not `0` — and a downstream `provisional == 0` mask silently drops it, while the sibling `is_play` on the line above is already guarded with `.fill_null("")`. Worse, an `open_ids` set holding only `None` is **truthy**, so the `is_in` branch runs with a needle that can never match: zero rows flagged while the feed still has unsettled plays. Neither is observed (0 null `playId`s across all 7,598 archived payloads), so this is hardening — `.fill_null(False)` plus discarding `None` from the set.

### Two things left for you

**The current-situation row is not inert under `enrich=True`.** It is the next state, so it also completes the **previous** play's lead-diff columns — measured, appending it flips 15 columns on exactly 1 row: `epa`, `qb_epa`, `rush_epa`, `pen_epa`, `qbr_epa`, `wpa`, `vegas_wpa`, `vegas_home_wpa`, the four `total_*_epa`/`total_*_rush_wpa`, `action_play`. That is arguably the right answer (nflfastR's `epa` *is* `ep(next state) − ep(this state)`), but the docstring's "every play-outcome column is null, so no consumer can mistake it for a snap that happened" reads as isolation, and that play is usually still `provisional` so the value moves on the next poll. I documented it; no test covers it.

**A `summary` without `timeouts` yields a scored-but-unscoreable situation row.** Dropping `summary.*.timeouts` leaves both timeout columns null and moves `wp` 0.110 → 0.070 and `vegas_wp` 0.159 → 0.031, with no warning. `test_current_situation_row`'s "enrichment inputs are all present" loop excludes both timeout columns, so it cannot catch this. If Shield ever omits them, the cheap fix is falling back to the last play's running timeouts when `last["posteam"] == posteam`. Unverified against the feed — left alone.

### Notes, not blockers

- **No bot has reviewed this PR.** sourcery-ai's only review is at `89acc648a` (not head) and its body is the rate-limit notice; the `Sourcery review` check is `skipped` at head; `reviewThreads` is empty. Nothing else has read this diff.
- `shield_nfl_pbp` and `is_final` are not reachable from `sportsdataverse.nfl`, so `generate.py`'s coverage scope (built from `dir(sportsdataverse.nfl)`, `generate.py:2196`) cannot see them — which is *why* the codegen check is green with no NFL reference-doc change. Correct as-is, but the PR's deliverable has no reference entry, and `coverage_allowlist.yaml` has zero `shield` entries, so hoisting these names later goes red unless the allowlist and `docs/docs/nfl/reference/additional.md` move in the same commit. `live.game_phase` is public-by-name, unexported, with no caller outside `live.py` (→ `_game_phase`), and `shield_nfl_pbp` is missing its `See Also:` block.
- `chore/codegen-post-531` (remote, 1 commit, **no open PR**) regenerates the identical `docs/docs/cfb/reference/additional.md`. This PR is the only path that lands it — delete the orphan branch after merge rather than opening a PR for it.
- A zero-drives INGAME payload leaves 90/91 rows with a null `posteam` while still emitting a valid situation row (pre-existing `build_pbp`, unobserved in the feed). HALFTIME flags 2 settled end-of-half plays `provisional` — the documented ceiling of the trailing-null heuristic, not a defect. `enrich=True` on a cold cache attempts four ~34 MB model downloads; pre-warm `models_dir` on the poller host.

### Gates

`pytest tests/nfl/shield_pbp/ tests/nfl/test_nfl_build.py` → 176 passed (172 at `1c311c9fa`). Four mutations on my own fixes, four caught. `ruff check` / `format --check` clean; `mypy` as CI runs it → Success, 420 files. Full offline suite as CI runs it → green. CI at `1c311c9fa` was all green (`codegen --check`, `ruff + mypy`, `pytest (offline)`, `release-manifest audit`; live + Sourcery skipped).

Full report: `background-research/2026-09-17-football-sources-program/reviews/sportsdataverse-py-536/review.md`.

Not merging — that's yours.

PR #540, opened 2026-09-18
feat(nfl): Shield → ESPN-summary adapter (source="shield" for NFLPlayProcess)

saiemgilani:
## What

`sportsdataverse/nfl/shield_pbp/to_espn_summary.py` — the **second projection of the one Shield
parse**. `shield_nfl_pbp` already turns a Shield `gamedetails` payload into nflverse-shape rows
(PR #528 + #536); this re-skins **those same rows** as an ESPN summary dict, so the unmodified
`NFLPlayProcess` runs on Shield data through `espn_nfl_pbp(summary=)`. Registered as
`source="shield"` in the dispatcher; ESPN stays the terminal fallback.

No edit to `nfl_pbp.py` internals. Rebased on #537.

```python
from sportsdataverse.nfl import shield_to_espn_summary, NFLPlayProcess

summary, notes = shield_to_espn_summary(game_detail, idmap_row)
proc = NFLPlayProcess(gameId=401872922, join_participants=False)
proc.espn_nfl_pbp(summary=summary)
result = proc.run_processing_pipeline()      # 533 play columns + 23 box sections
```

or, through the switch contract:

```python
_process_game("nfl", 401872922, source="shield", idmap_row=row)   # provenance["served"] == "shield"
```

## The re-skin, and how each rule was decided

Every rule below was **measured against the real ESPN feed**, not assumed.

| piece | rule | evidence |
|---|---|---|
| play id | `{espn_event_id}{shield playId}` | the nfl-raw crosswalk's identity |
| drive grouping, `drive.id` | `{espn_event_id}{1-based index}`; a play joins the last drive that started at or before it | **6,323/6,323** crosswalked 2025 plays land in ESPN's own drive |
| `type.id` / `.text` / `.abbreviation` | 32-entry ESPN catalogue enumerated from **582** ESPN NFL summaries, resolved from the parser's stat flags | GOP bracket-reads `type.id` and `type.abbreviation` (`app.py:204-208`) |
| `yardLine` | absolute, from the **home** goal line; `yardsToEndzone` relative to `start.team` | verified on every crosswalked 2025 game |
| kickoff `start.team` | the **kicking** team, which the processor flips via `kickoff_vec` | ESPN's own convention |
| clock | `M:SS`, never zero-padded; a **scoring** play carries the *next* row's snap clock | agreement 0.27 → 0.99 on the first fix |
[...]

PR #541, opened 2026-09-18
feat(cfb): Yahoo → ESPN-summary adapter (source="yahoo")

saiemgilani:
Adds `sportsdataverse/cfb/yahoo_pbp/` — the Yahoo → ESPN-summary adapter for the football
source-switch contract (#525), registered as **`source="yahoo"` for the CFB**. The consumer is the
**unmodified** `CFBPlayProcess` through `espn_cfb_pbp(summary=)`; ESPN stays the terminal fallback and
nothing else in `football/sources/` changes (the dispatch diff is one `_ADAPTER_MODULES` entry plus the
docstring sentence that lists the registered sources — a CBS NFL adapter is in flight against the same
module).

Everything in the package is `_`-prefixed and nothing is re-exported from `sportsdataverse.cfb`, so
**no codegen and no reference-doc regeneration is involved** — the same choice `shield_pbp` made.

Full evidence: `background-research/2026-09-17-football-sources-program/s2-yahoo-cfb/FINDINGS.md`
(+ `g1_report.md`, `g1_divergences.txt`, `g2_g3.md`, cached payloads and processed frames).

## How the Yahoo game id is resolved

Yahoo is the only CFB source whose game id needs no stored map:
`ncaaf.g.{US-Eastern kickoff date YYYYMMDD}{home team's Yahoo number:04d}`. The cascade, stamped in
`native_ids["yahoo_id_resolved_by"]`:

1. `idmap` — the row's stored `yahoo_game_id` (sdv-db #68 already computes and stores one per CFB game);
2. `computed_idmap_team` / `computed` — the formula, taking the Yahoo number from the row's
   `home_team.yahoo_team_id` if it has one, else from a committed 261-team ESPN→Yahoo table
   (`yahoo_pbp/teams.py`, built once from `cfb_teams_crosswalk(providers=("espn","yahoo"))` — it cannot be
   built at request time, because its ESPN leg is built on ESPN);
3. `cfb_schedule_crosswalk` — the published `cfb_crosswalk` release asset (2014+), cached per process; the
   only leg that works with **no id-map row at all**, since the formula needs a kickoff date;
4. otherwise `SourceUnavailable`. **An id is never invented.**
5. Resolution is skipped entirely when a caller injects the payload, which is what makes the tests offline
   by construction.

Verified this session: **31/31** ids reproduced for the captured 2024–2026 FBS finals, and for both
FCS-hosted probes.

### G1 — finals parity vs ESPN, **30 FBS games 2024–2026**, ≤1 req/2 s

Both paths through the unmodified processor with **one** closing line injected into both, so the
comparison isolates the feed rather than the odds.

Yahoo has no ESPN play ids, so the join is on game state — and the repo's `parity.STATE_KEY` could not be
used: it contains `clock.displayValue`, and **ESPN's CFB feed repeats one clock across several consecutive
plays** (15:00 ×4, then 13:52 ×3) while Yahoo stamps every play. The clocks agree on **9.5%** of paired
rows. Key used: `(period, start.down, start.distance, start.yardsToEndzone, start.pos_team.id)`, with
clock stoppages and any state key that is not unique within the game dropped from **both** sides.

| | value |
|---|---|
| games | 30 (2024: 11, 2025: 12, 2026: 7) |
| plays | ESPN 5,091 / Yahoo 4,942 — **ratio 0.971** |
| **join rate** | **3,799 / 4,004 = 94.9%** of ESPN's unambiguous snaps |

| column | exact | | metric | r (pooled) | mad | target |
|---|---|---|---|---|---|---|
| `start.yardLine` | **0.9968** | | `EP_start` | **0.9992** | 0.031 | ≥ .998 ✅ |
| `end.pos_team.id` | **0.9963** | | `wp_before` | **0.9996** | 0.004 | ≥ .998 ✅ |
| `pos_score_diff_start` | 0.9913 | | `wp_after` | **0.9985** | 0.005 | ≥ .998 ✅ |
| `type.text` / `type.id` | 0.9829 / 0.9760 | | `EP_end` | 0.9883 | 0.072 | ≥ .998 ❌ |
| `rush` / `pass` | 0.9929 / 0.9961 | | `EPA` | 0.9570 | 0.083 | ≥ .99 ❌ |
| `penalty_flag` / `td_play` | 0.9929 / 0.9939 | | | | | |
| `turnover_vec` | 0.9982 | | | | | |
| `statYardage` | 0.9497 | | | | | |
| `clock.displayValue` | **0.0953** | | | | | |

**Where the gap is.** `end.yardsToEndzone` agrees on **97.66%** of paired rows, and splitting on that one
column accounts for the whole shortfall:

| subset | n | EPA r | EPA mad | EP_end r |
|---|---|---|---|---|
| same type **and** same end spot | 3,637 | **0.9838** | 0.051 | **0.9974** |
| end spot differs | 89 (2.3%) | 0.6733 | 1.099 | 0.8216 |

57 of 3,799 rows (1.5%) move EPA by more than 1.0. Causes, with examples in `g1_divergences.txt`:

1. **ESPN folds what Yahoo splits** — ESPN writes a punt and the penalty enforced on it as one row, Yahoo
   as two, so the next-snap end state is the pre-enforcement spot. Structural; the two feeds describe the
   same football differently.
2. **The feeds genuinely disagree about the play** — "pass incomplete to Jaron Tibbs" against "passed to
   Jaron Tibbs for 28 yard gain"; `Fumble Recovery (Own)` against `(Opponent)`. One is wrong and it is not
   always Yahoo.
3. **ESPN's own type vocabulary is not stable across seasons** (a returned kickoff is
   `Kickoff Return (Offense)` in 2024–25 and `Kickoff` in 2026; a returned punt is `Punt` in 2024 and
   `Punt Return` in 2025–26), which is most of the 2.4% `type.id` disagreement. The adapter emits the
   modern, era-stable pair; both collapse to the same `PLAY_TYPE_CANONICAL` and neither moves EP or EPA.
4. `statYardage` 0.9497 — ESPN books a kick's **return** yards where Yahoo's `yards` is 0. That feeds
   `yds_punted` / `yds_kickoff`, already in `KNOWN_LOSSY[("cfb","yahoo")]`, not EP/EPA.

**G1 partially met**: EP-before and both WP columns clear ≥ .998; EPA clears .98 on the 95.7% of rows where
both feeds agree what the play ended as and is 0.957 pooled. The residual is ESPN-vs-Yahoo disagreement and
ESPN's stale clock, neither of which an adapter change can close.

### G2 — in-progress payloads ✅

No FBS game was live while this ran, so the snapshots are **truncated real finals** (payload cut at a play
id, drive chart cut with it, status set to `IN_PROGRESS`). **8 snapshots** across two games, 2 → 146 plays:
every one puts the open drive in `drives.current` with a synthetic end and nulled outcome columns
(`result` = "In Progress", `isScore` False, `status.type.completed` False, no `winner`), passes
`_validate_summary`, and processes. The first two snapshots are the case the #540 review caught: a live
game whose only drive is the open one has an empty `drives.previous`, so the play-presence guard reads
**both** groupings.

### G3 — FCS-hosted = the contract's "no coverage" failure ✅

Elon at Rhode Island (`401866628` / `ncaaf.g.202609120206`). Yahoo answers **HTTP 200** with a real game
object — status `FINAL`, score 10–31, a win-probability stub — and zero plays and zero drives:

```
SourceUnavailable: cfb 401866628: yahoo game ncaaf.g.202609120206 carries no play-by-play
                   (FCS-hosted, or before Yahoo's 2014 play floor)
```

Detected by **shape**, and checked before anything about the id-map row, so the message is the same for
every FCS-hosted game and dispatch continues down `SOURCE_ORDER`. It is deliberately a different failure
from "no shangrila game object", which is what a rate limit's 23-byte `text/html` body produces and which
**is** worth retrying (`dl_utils.download` already retries 429 with `Retry-After`).

## Tests — 17, **zero network**

Three committed trimmed payloads (9.6 KB / 11 KB / 1.3 KB gzipped) plus ESPN summaries already in
`tests/cfb/fixtures`. Both mutations verified red:

| mutation | test that goes red |
|---|---|
| `target = emitted[-1]` instead of the newest touchdown | `test_pat_anchors_on_its_touchdown_across_an_intervening_row` — LOU @ MISS carries a false start between a touchdown and its PAT and a timeout before a two-point try. **The same mutation passes on a fixture whose tries are all adjacent**, which is how this defect shipped green in #540 |
| drop the `_has_plays` guard | `test_an_fcs_hosted_game_reports_no_coverage_...` + `test_a_payload_with_no_plays_at_all_hands_over` |

The #540 review's six majors were used as the checklist: PAT fold target, play-presence guard across
`current`/`previous`, club codes in the header (here: ESPN team ids from the id-map row mapped onto the
payload's own Yahoo ids, and a non-empty mascot so a `Timeout` row does not charge both clubs), a
Data-API-shaped id-map row (no `home_team` sub-row needed), no network in tests, and the scoring end state
on defensive touchdowns (credited to `scoringPlayInfo.scoringTeamId`, or a pick six's end spot flips 100
yards).

## Known limits

FCS-hosted games and pre-2014 plays are not served in any era; lineups are empty before 2017 so older text
keeps Yahoo's `[ncaaf.p.N]` placeholders (surfaced as a note); `teams.py` is a static snapshot, where a new
FBS member is a *missing key* → hand-over, never a wrong id.

## Summary by Sourcery

Add Yahoo CFB play-by-play failover support through an ESPN-compatible adapter and register it as the Yahoo football source.

New Features:
- Add a Yahoo college-football play-by-play adapter that produces ESPN-summary-compatible data for the existing CFB processing pipeline.
- Support Yahoo game ID resolution through stored mappings, deterministic team/date computation, and the published schedule crosswalk.
- Handle live payloads, unsupported no-coverage games, score conversions, drive state, and source fall-through behavior.

Bug Fixes:
- Prevent unsupported or malformed Yahoo responses from being treated as empty valid games.
- Preserve correct touchdown/PAT scoring, defensive scoring state, and open-drive play presence across edge cases.

Enhancements:
- Register Yahoo as a CFB football source while keeping the existing processor and ESPN fallback unchanged.

Tests:
- Add offline fixture-based coverage for finals parity, in-progress games, unsupported FCS coverage, ID resolution, dispatch behavior, and malformed or mismatched payloads.

## Summary by CodeRabbit

* **New Features**
  * Added Yahoo as a fallback play-by-play source for college football games.
  * Converts Yahoo play-by-play into the standard game summary format, including drives, scoring plays, team details, game status, and available odds.
  * Resolves Yahoo game identifiers using stored mappings, game details, team information, or published crosswalk data.

* **Bug Fixes**
  * Gracefully handles unavailable, incomplete, mismatched, rate-limited, or play-free Yahoo game data.


saiemgilani (commented):
Scoped review of the Yahoo → ESPN-summary adapter: the six major classes from the Shield review (#540) re-checked against this adapter with a probe each, an independent G1 reproduction with the EPA attribution tested by removal, the clock-free join spot-checked, the no-coverage guard mutated, the team table, and the polars/docstring lenses. Everything below was run offline in a bubblewrap sandbox against a `uv sync --frozen` venv in the review worktree.

**The projection is right, and the numbers reproduce exactly.** Re-running G1 from scratch over all 30 cached games from the raw payloads: 3,799 paired, EPA r .9570, EP_start .9992, EP_end .9883, wp_before .9996, wp_after .9985, and 89 rows where the two feeds put the ball in different places. Same to four decimals.

**But the 89 are not all the feeds.** Classifying every one of them by what follows it in the Yahoo feed:

```
end spot agrees 3710/3799 = 0.9766; differs 89
    33  other (made-FG convention, punt/penalty folding, turnover disagreements)
    27  LAST_PLAY_OF_FEED (end := own start)        <-- the adapter, not the feed
    13  FEED_DISAGREES_on_yardage
    12  FEED_DISAGREES_on_play_type
     4  next row Penalty (other)
```

`_fill_end_state` took `next(<next non-stoppage snap>, play["start"])`. The fallback fires on the last snap of every game and, on the live path, on **the newest snap of every poll** — the row Game on Paper renders at the top of the page. It says the ball never moved:

```
espn   Rush  d=2 dist=8 sYTG=90 -> eYTG=71  EPA=1.02   "Kaytron Allen run for 19 yds to the PSU 29"
yahoo  Rush  d=2 dist=8 sYTG=90 -> eYTG=90  EPA=0.53   "Kaytron Allen rushed for 19 yard gain"
```

27 of the 30 games carry one. Fixed by deriving the end from the play's own yardage on a plain scrimmage snap; a turnover or a kick is left alone rather than guessed at, because its yardage says nothing about where the next team starts.

Re-running G1 with the fix: **89 → 69** divergent rows, the adapter's share **27 → 6** (the 6 left are exactly those turnovers and kicks). Pooled EPA r moves **.9570 → .9571**, i.e. not at all — these rows carry small |dEPA| individually and the pooled correlation is dominated by the handful of high-|dEPA| rows where the feeds report different football. That is the useful result: the .957 residual is now *demonstrably* the feeds, not the adapter.

**The made-field-goal convention is not ESPN's.** The branch writes `yardsToEndzone 65` and the docstring calls that "ESPN's measured convention … the ensuing kickoff spot". ESPN's own captured payloads:

```
401628414 start ytg  6 -> RAW ESPN end {down:-1, distance:-1, yardLine:94, yardsToEndzone: 6}
401628414 start ytg 26 -> RAW ESPN end {down:-1, distance:-1, yardLine:74, yardsToEndzone:26}
401628414 start ytg 36 -> RAW ESPN end {down:-1, distance:-1, yardLine:64, yardsToEndzone:36}
```

`down -1 / distance -1` is ESPN's; `65` is not. EPA is unaffected (|dEPA| ≈ 0.00 — the processor scores a made kick from its outcome); `end.yardsToEndzone`, `end.yardLine` and `EP_end` were not. Fixed to the measured convention.

**The wrong-game guard.** CodeRabbit's open Major thread ([r4050516952](https://github.com/sportsdataverse/sportsdataverse-py/pull/541#discussion_r4050516952)) is half right, and the half that is right is the reachable one: the injected-payload case is the caller's own contract (the same affordance ESPN's `payloads=` has), but on the **fetch** path the Yahoo id is a *formula*, so a stale team number or a moved kickoff resolves to a real game that is not this one — and nothing checked. Yahoo echoes the id it served, so the mismatch is checkable in three lines; it is now a hand-over instead of another game's plays filed under this ESPN event id.

Also fixed: `_crosswalk_yahoo_id` raised `ColumnNotFoundError` out of a function documented "never raises" when the asset renames `espn_game_id`; a non-mapping payload raised `TypeError` instead of the documented `ValueError`; the `Returns`/`Raises` tables listed 4 of the 5 notes and 1 of the 2 `ValueError`s the code emits; and the open drive's outcome is now nulled where it is popped into `drives.current`, since the drive that becomes `current` is the last one *with plays*, not always the last entry in the chart.

### What held up

| #540 class | here |
|---|---|
| PAT folds onto a stoppage row | handled — anchors on the newest touchdown, mutation-verified |
| play-presence guard misses `drives.current` | handled — the guard reads both groupings |
| club codes reach ESPN's header | not reachable — team **ids** come from the id-map row; no abbreviation table to disagree with |
| Data-API-shaped rows invent values | not reachable — the adapter reads only `GAME_SCHEMA` columns |
| network in tests | clean — 20 tests, `--unshare-net`, 20 passed |
| scoring end state on defensive TDs | handled — credits `scoringPlayInfo.scoringTeamId`, so a pick six does not flip 100 yards |

Two more that held: the no-coverage guard is mutation-verified red (`_has_plays → True` ⇒ 2 tests fail), and because `FCS_ROW` deliberately carries no team ids, `match="no play-by-play"` also pins the **ordering** — moving the shape check after the id-map check turns it red too. And `teams.py` degrades safely: 261 entries with **no Yahoo number claimed by two ESPN teams** (a name-match collision is what would have produced a wrong id), and an unknown ESPN id walks the whole cascade to `SourceUnavailable`, never a guess.

The clock-free state key is sound: 20 pairs spot-checked by play text, 20/20 the same play, 0 mis-paired.

### On EPA .957

**Accept it for a failover source, with the number and its composition written into the PR body and into dispatch's provenance.** The end-spot fix above had to happen on its own merits, but it is not a material share of the gap — removing it left r at .9571. EP_start (.9992) and both win-probability columns (.9996 / .9985) clear .998 pooled over 30 games, and those are what a page renders. What the residual is, having classified all 69 remaining rows: ESPN folding what Yahoo splits (a punt and the penalty enforced on it as one row against two) and the two feeds genuinely reporting different football (a completion against an incompletion, own-fumble against opponent-fumble) — neither of which an adapter change can close, and on which ESPN is not always the one that is right. This source is never first in `SOURCE_ORDER`; it serves a game only when ESPN has already failed, and the alternative to .957 is no game at all.

Three new tests, each verified red with its fix mutated out. `ruff`/`mypy` clean. Full review with every ledger: `background-research/2026-09-17-football-sources-program/reviews/sportsdataverse-py-541/review.md`.

mifuha:
I confirmed the equivalent failed-fourth-down case remains in the Yahoo adapter on `main` at commit `5fb17d39c28848874e01e9ec16cdd9552e365446`.

I reproduced it using the committed OSU–Texas Yahoo fixture (`401856682`, play 43).

The play is 4th-and-2 at the Texas 13, with Ohio State gaining 1 yard. When the live snapshot ends on that play, Yahoo currently produces:
```text
end.team.id = 194        # Ohio State
end.yardsToEndzone = 12
end.down = 4
end.distance = 2
```

and those values also survive into the processed frame.

With the next snap present, the same play correctly ends as:
```text
end.team.id = 251        # Texas
end.yardsToEndzone = 88
end.down = 1
end.distance = 10
```

`end.yardLine` stays correctly at the absolute 12 in both cases.

This looks like the Yahoo equivalent of the failed-fourth-down handling identified/fixed for CBS in #545 (and subsequently Fox). I also checked a second case with the home offense and got the same result.

I haven't opened a separate issue or made a competing patch since the defect class has already been discussed. Happy to put together the small Yahoo regression/fix if that would be useful.

PR #553, opened 2026-09-19
feat(validation): per-game validation gate — validate_game + GameReport (V1)

saiemgilani:
## What

`sportsdataverse.validation.validate_game(frame, league, *, header=None, source="espn") -> GameReport` —
one pure, offline function every consumer (a release build stage, the Data API, Game on Paper) can call
on a processed game. V1 of `ClaudeCowork/plans/2026-09-18-data-integrity-and-live-monitoring.md`.
V2/V3 are not in scope.

**`tools/` does not ship in the wheel**, so the rules had to move, not be copied:

| from | to |
|---|---|
| `tools/validation/checks/pbp_invariants.py` | `sportsdataverse/validation/pbp_invariants.py` |
| `tools/validation/findings.py` | `sportsdataverse/validation/findings.py` |
| `tools/validation/thresholds.yaml` | `sportsdataverse/validation/thresholds.yaml` (packaged data) |

`tools/validation` keeps the offline runner and imports from the package. **No shim modules** — every
importer is repointed, so each symbol has exactly one import path and there is one rule set, not two.

`GameReport` carries `game_id, league, source, season, processing_version, n_rows, ok, errors,
warnings, counts_by_rule`, with `to_dict()` (stable key order, JSON-serialisable) and `to_row()` for
V2's `<family>_qa_{season}` asset. `Finding` is `(rule_id, severity, n_rows, n_checked,
sample_row_ids, message)`.

**Rules carry scope, not just severity.** `RULE_SCOPE` narrows a rule per league, per source and per
era, and every entry cites the sweep finding that measured it: run3's R2 (ESPN emits >3 team-named
timeout rows per half in modern CFB), R3 (goal-to-go text that contradicts the corrected spot), R4 /
C41 (pre-2015 feeds re-enter score rows after the scoring play), R5 (NFL touchdown yardage is ESPN's
own arithmetic), O10/O12/O13, and the CBS gate's finding that ESPN's own `end.distance` matches the
next `start.distance` on only .30–.78 of adjacent pairs 2018–24. Counts are kept whatever the
severity, so scoping never hides a rule; a rule outside its `era_scope` is demoted to `warn`, and a
rule that only judges ESPN's feed is skipped entirely for an adapted source.

**Wiring.** `NFLPlayProcess.run_processing_pipeline(validate=False)` and `CFBPlayProcess` likewise
attach the report under the game's `"validation"` key when True; `football.sources.dispatch` always
runs it and merges it into the existing provenance beside the contract report (guarded — a rule that
raises costs the caller a provenance field, never the game).

## Rules

**89 rules: 52 `error`, 27 `warn`, 10 `info`.** Full table with per-rule scope and the reason for each
scope: `background-research/2026-09-17-football-sources-program/v1-validation/rule_table.md`.

New in this PR (the aggregations the spec asked for that did not already exist):

| id | group | severity | assertion |
|---|---|---|---|
| `period.monotone` | 11 | error | the period never decreases in row order |
| `clock.monotone_within_period` | 11 | warn | the clock never runs backwards inside a period |
| `type.null_id` | 11 | error (warn pre-2010) | every live play carries ESPN's `type.id` |
| `type.null_abbreviation` | 11 | warn | ESPN itself emits null on Sack / Pass Incompletion |
| `ytg.continuity_next_snap` | 2 | warn | end state == the next snap's start **across** a drive boundary (`100 - end` on a turnover); `ytg.continuity` only looks within one drive |
| `epa.team_sum_matches_box` | 12 | error | Σ row EPA over a team's scrimmage plays == `advBoxScore` `EPA_overall_off` |
| `plays.kickoff_count_matches_scores` | 12 | warn | kickoffs == scores that restart play + one per half |
| `drive.count_matches_feed` | 12 | warn | one processed drive per drive in the feed |

Already present, so not re-implemented: final score vs header, WPA sums to the result, box parity vs
the source box, play count vs the feed, possession alternation.

## Sweep identity (behaviour preservation)

Same 105-game slice of run2's manifest (NFL 2023–25 = 60, CFB 2024–26 = 45), **both sides pinned to
the same base commit**, full processing both times, 3 workers, offline:

```
games: before=105 after=105 same_keys=True
rules: before=81 after=89 shared=81 differing=0
only_before: []
only_after : clock.monotone_within_period, drive.count_matches_feed, epa.team_sum_matches_box,
             period.monotone, plays.kickoff_count_matches_scores, type.null_abbreviation,
             type.null_id, ytg.continuity_next_snap
games differing on shared rules: 0
```

Every pre-existing rule's `n_checked` **and** `n_violations` is identical on every game; the only
difference is the eight new rule ids. Per-rule totals in `v1-validation/sweep_identity.md`.

> Worth recording: the first before/after pair differed on 4 NFL 2025 games. That was not the
> refactor — `main` moved under the runs (#546, N37 strip-sack). A sweep-identity claim only means
> something when both sides are pinned to the same commit.

## Gate verdict on main today — what V2 inherits

| league | games with no `error` finding (same 105) |
|---|---|
| nfl | 44/60 (73%) |
| cfb | 21/45 (47%) |

The rules that fire are the ledger's **open** defects, not new ones: `score.delta_value` (O2),
`ep.start_range` / `ep.end_range_non_scoring`, `score.monotone`, `poss.kickoff_receiver_gets_ball`,
`wp.home_wp_after_complemented`. **V2 has to plan against this**: the spec's straw-man "block a
publish above 0.5% of games with an error" would block every season today, so V2 should start
report-only (publish the `*_qa_*` asset, do not fail the build) and ratchet as those items close.
Full table: `v1-validation/gate_rates.md`.

## Tests, codegen, gates

* `tests/validation/test_validate_game.py` — report contract, era/source scoping, a drift guard that
  `RULE_SCOPE` cannot name a rule the table does not produce, and the gate asserted on **six
  committed real-game fixtures** (3 per league) processed offline. `tests/contracts/test_pbp_invariants.py`
  gains a fires/does-not-fire pair for each new rule. 0 network.
* `tests/validation tests/football/sources tests/contracts`: **336 passed, 1 skipped** on the rebased tip.
* `tools/codegen/generate.py` then `--check`: **all generated files current**. `validate_game` is
  re-exported at the package top level so the autodoc documents it on `reference/python-helpers`, and
  the processors' `validate=` keyword lands on each league's `reference/additional`. The report's
  fields are documented as dataclass tables in the docstrings (the `dispatch.py` convention), so no
  `manual_column_descriptions.yaml` entry is needed.
* `ruff check` / `ruff format --check` clean; `mypy` clean, and all three new modules are added to the
  strict-typing ratchet.

## Follow-up recorded, not built: V1b (advBoxScore reconciliation)

Saiem asked mid-build for the derived box tables to be covered too. Assessed and deliberately not
stretched into this PR. The hook already exists — `validate_game(..., box=...)` takes the
`advBoxScore` dict today and `epa.team_sum_matches_box` is the family's first rule — so V1b is a
one-day follow-up over 21 sections with 7 proposed rules (`box.team_totals_match_plays`,
`box.player_sums_match_team`, `box.rates_recompute_from_counts`, `box.sections_mirror_off_def`,
`box.turnovers_match_flags`, `box.drives_match_drive_rows`, `box.usage_shares_sum_to_one`). Exact
list, scope and hook point: `v1-validation/FINDINGS.md` §V1b.

## Evidence

`background-research/2026-09-17-football-sources-program/v1-validation/` — `FINDINGS.md`,
`sweep_identity.md`, `rule_table.md`, `gate_rates.md`, and both sweep output trees.

## Summary by Sourcery

Ship a reusable per-game football data-integrity gate and integrate its structured verdicts across processors and source dispatch.

New Features:
- Add the packaged, offline `validate_game` API with structured `GameReport` and `Finding` results for per-game data-integrity checks.
- Add scoped play-order, type-identity, cross-drive continuity, EPA reconciliation, kickoff, and drive-count validation rules.
- Attach validation reports through optional processor validation and automatic source-dispatch provenance integration.

Enhancements:
- Move the shared validation rules and findings into the installable `sportsdataverse.validation` package so production consumers and offline tooling use one rule set.
- Support league, source, and era-aware rule severity and preserve rule counts across scoping decisions.

Build:
- Package validation threshold data in the distribution and extend strict typing coverage to the new validation modules.

Documentation:
- Document the validation API and processor validation options in the generated reference documentation.

Tests:
- Add report-contract, scoping, new-rule, processor-integration, and offline real-game fixture tests.

Chores:
- Repoint validation tooling and contract tests to the packaged validation modules.

## Summary by CodeRabbit

* **New Features**
  * Added offline per-game validation for NFL and college football play-by-play data, with structured reports covering errors, warnings, and rule counts.
  * Added a public validation helper available from the main package API.
  * Processing pipelines can optionally attach validation reports to returned games; processed game provenance now includes validation results.
  * Expanded checks to cover play ordering, continuity, aggregations, scoring, kickoff counts, and drive counts.

* **Documentation**
  * Added usage and parameter documentation for validation and pipeline options, including selective report returns.


saiemgilani (commented):
## Review — `feat/v1-validation-gate` @ `7dbb4841` → fixes pushed to the branch

The move is clean and the identity claim holds. I re-ran the pinned before/after myself and then went
after the new rules on real games; seven things needed fixing and I've pushed them to the branch.
Full report with every command and output: `background-research/2026-09-17-football-sources-program/reviews/sportsdataverse-py-553/review.md`.

### Verified first

**Sweep identity reproduces exactly.** 10 games of your manifest, both sides pinned to `5d1688ed2`
(`origin/main` tools-side in a fresh detached worktree, this branch package-side, separate
`uv sync --frozen` venvs, both offline):

```
same game keys: True 10 10
games/rules differing on shared rules: 0
per-rule total mismatches: []
only_after: the 8 new rule ids
```

`pbp_invariant_sweep.py`'s per-game JSON keys are byte-for-byte the same shape as `s1-sweep/run3/games/**`,
so `FINDINGS_run3.md`'s comparisons stay reproducible.

**`validate=` is inert.** 3 real games, frames written to uncompressed parquet and SHA-256'd:
`validate=False` == `validate=True` == `main` on all three; the only change to the returned dict is the
`"validation"` key.

**Gate rate confirmed.** My own 27-game slice: **nfl 10/14 (71%), cfb 6/13 (46%)**, same top failing rules.

**Packaging.** `uv build --wheel` ships `sportsdataverse/validation/thresholds.yaml`; `tools/` is absent;
`sportsdataverse.validation` costs 12.6 ms at import with no file read or polars scan.

### Fixed in-branch

1. **`ruff format --check` was red at head.** CI's `ruff + mypy` job on `7dbb4841` is `failure` — one
   missing blank line in `sportsdataverse/__init__.py`. The body says the format check is clean.
2. **`plays.kickoff_count_matches_scores` made every overtime game a false positive.** `expected` counted
   OT scores (which restart from the 25, no kickoff) and PAT / two-point / defensive-conversion rows
   (which ride the touchdown's kickoff). Its only two firings in 105 games were both OT: `401628439`
   11 kickoffs vs 19 "expected", `401858224` 11 vs 16. After the fix both land inside tolerance and the
   rule fires **0/105**. Regression test pins the OT frame.
3. **`period.monotone` shipped at `error` but measures ESPN's row order.** Both firings are feed defects —
   `401628439`'s raw `drives.previous` carries a period-2 row after period 6, and `401752913`'s
   interception-return row sorts by play id into the next drive's block. That is the same class `RULE_SCOPE`
   already demotes for R2/R3/O10, so it's now `warn` with the two game ids in the note.
4. **`epa.team_sum_matches_box` read a failed id join as a clean pass.** `rows[int(r["pos_team"])]`
   swallowed a non-int id and `want is None: continue` skipped unmatched teams, while `n_checked` counted
   frame teams. A renamed box key — or `pos_team` arriving as `"194.0"` — gave `n_checked=2, n_violations=0`
   on an **error**-severity gate. Its `0/210` across the sweep was consistent with "nothing matched".
   An unreconciled team is now a violation and `n_checked` counts matched teams; three mutation cases added.
   Worth carrying into V1b, since the `box.*` family replicates this shape ten more times.
5. **`down.scrimmage_distance_ge_1` was switched off for every adapted source.** The rule reads only
   `start.distance` — nothing ESPN-specific — so `sources=("espn",)` meant a cbs/yahoo/fox/shield/ncaa
   adapter emitting distance < 1 was invisible (and absent from `counts_by_rule` too). O13's evidence is a
   per-league claim and `era_scope={"nfl": …}` already carries it. Source restriction dropped.
6. **The gate failed open on an unresolved season.** `severity_for` collapsed `window is None`,
   `season is None` and "outside the window" into one `warn`, so a frame with no resolvable season reported
   `ok=True` with every era-scoped error rule demoted. Not reachable today (`header.season.year` is a
   required contract field and the frame carries `season`), but a gate shouldn't answer green when it
   couldn't decide. Unresolved season now keeps the rule's severity.
7. Minors: the guarded `dispatch` path returned `{"error": …}` with no `ok` key, so
   `report.get("ok", True)` read a crashed gate as a pass → `{"ok": False, …}`; the kickoff rule's
   `n_checked` was the literal `1`, which made the harness' `assert n_checked > 0` guard vacuous for it;
   `validate_game` had no `Raises:` for its `ValueError`; and the `validate=` docstring promised the report
   "on the returned game", which isn't true when `return_keys` was given at construction (docs regenerated).

7 mutations run, 7 caught (including one each on the two new fixes).

### Two things left for you

**The gate-verdict claim is overstated.** "The rules that fire are the ledger's **open** defects, not new
ones" — only `score.delta_value` (O2) is a ledger item. `ep.start_range`, `ep.end_range_non_scoring`,
`poss.punt_changes_possession`, `flags.completion_on_incomplete_text`, `flags.no_play_yardage_credited`
and `flags.punt_type_without_punt` appear nowhere in `LEDGER.md`, and `period.monotone` is this PR's own
new rule. V2's threshold plan is being built on that sentence, so it's worth correcting in the body.

**And the two biggest blockers have one root cause, which is new.** `ep.start_range` and
`ep.end_range_non_scoring` fire on 10/105 games each — and on *the same single row* every time:

```
{"type.text": "Official Timeout", "text": "Official Timeout at 02:27.", "EP_start": 8.0, "EP_end": 8.0, "EPA": 0.0}
```

`EP` = 8.0 on `Official Timeout` rows, outside the documented [-7, 7] band. One processor defect, 20 of
the error firings, ~10% of games. That's the cheapest lever on the gate rate before V2 sets any threshold —
it belongs in the ledger as a new item.

### Smaller notes (not blocking)

* Six `RULE_SCOPE` entries are no-ops — the rule table already gives them `warn`. Harmless, worth a comment.
* Three new rules set `Severity.WARN` at the call site rather than in `RULE_SCOPE`, so they carry no cited
  measurement; the discipline `RULE_SCOPE` enforces doesn't reach them.
* `ytg.continuity_next_snap`'s note cites the CBS gate's `end.distance`-vs-`start.distance` rate (.30–.78),
  but the rule compares `yardsToEndzone`, whose own measured rate is cfb .034 / nfl .048.
* `getattr(self, "source", "espn")` in both processor hooks is a dead branch — neither processor ever
  assigns `self.source`, so the hook can only ever report `espn`.
* `test_real_games_pass`'s six fixtures were selected *because* they pass; one committed fixture with a
  pinned `failed_rule_ids` would give the suite a red side.
* `drive.count_matches_feed` reports `n_violations ∈ {0,1}` against `n_checked = len(feed_drives)`, so
  losing 21 of 22 drives shows as a 4.5% rate.
* `sportsdataverse/__init__.py`'s comment says "0.1.5+"; the wheel builds `0.1.4`.

All PR code ran inside the review sandbox (bwrap, `--unshare-net`, `/root` and `/etc/sdv-db` masked).
0 network. `ruff` / `ruff format` / `mypy` / targeted tests green at the fixed head.

